PR #3, opened 2024-09-28
Created a new SportCard component and updated build.gradle for preview

kelsiliul:
image
- Created a new SportCard component 
- Updated the build.gradle file with necessary dependencies and configurations


app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -0,0 +1,137 @@
...
+    date: String,
+    location: String,
+    sportName: String,
+    sportIcon: Painter
+) {
+    Card(
+        modifier = Modifier
+            .padding(16.dp)
+            .fillMaxWidth(),
+        colors = CardDefaults.cardColors(containerColor = Color.White)
```
thisjustin123, line 39: One tip to boost your component game:

Take in a `modifier: Modifier = Modifier` as input to the component, and instead of doing `modifier = Modifier` on the `Card`, work off of the `modifier` you take in as argument. This allows much more customization of the component by people using it.

Also, I think you should take off the `.padding(16.dp)` here. People calling the `SportCard()` can then use that `modifier` argument to add padding themselves if needed. The fill max width is ok to keep.

app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -0,0 +1,137 @@
...
+    Column {
+        SportCard(
+            teamLogo = painterResource(id = R.drawable.penn_logo),
+            team = "Penn",
+            date = "5/20/2024",
+            location = "U. Pennsylvania",
+            sportName = "Baseball",
+            sportIcon = painterResource(id = R.drawable.ic_baseball)
+        )
+    }
```
thisjustin123, line 136: Love it! Argument choice is really nice, and the preview looks great. I like that you're taking in the painters as parameters.

thisjustin123 (changes requested):
One small thing; plz add the `modifier` argument!

app/src/main/java/com/cornellappdev/score/components/SportCard.kt
kelsiliul: Addnmodifier as parameter and padding.

app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -25,6 +25,7 @@ import com.cornellappdev.score.theme.Style.universityText
 
 @Composable
 fun SportCard(
+    modifier: Modifier = Modifier,  // Add modifier argument
     teamLogo: Painter,
     team: String,
     date: String,
```
thisjustin123, line 31: One last thing; please move to be the **last** argument in the list!

thisjustin123 (approved):
Please fix that last thing and then it looks good to merge! Really great stuff!!

PR #5, opened 2024-10-03
Implement sport selector header

helenjb:
Screenshot 2024-10-03 at 16 21 48

- Implement UI for sport selector sticky header
- When gender or sport is selected, that option is colored red, and previously selected option is colored to gray


app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+    val sports = listOf(
+        Triple("All", painterResource(id = R.drawable.ic_all), painterResource(id = R.drawable.ic_all_filled)),
+        Triple("Baseball", painterResource(id = R.drawable.ic_baseball), painterResource(id = R.drawable.ic_baseball_filled)),
+        Triple("Basketball", painterResource(id = R.drawable.ic_basketball), painterResource(id = R.drawable.ic_basketball_filled)),
+        Triple("Cross Country", painterResource(id = R.drawable.ic_cross_country), painterResource(id = R.drawable.ic_cross_country_filled)),
+        Triple("Football", painterResource(id = R.drawable.ic_football), painterResource(id = R.drawable.ic_football_filled))
+    )
+
+    SportSelectorHeader(sports, {}, {})
+}
```
thisjustin123, line 50: Please put at the bottom of the file!

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+
+@Composable
+fun SportSelectorHeader(sports: List<Triple<String, Painter, Painter>>,
```
thisjustin123, line 53: I do like the idea of some data structure to pipeline in a string and two relevant painters. It's just a bit hard to read, it just being a triple.

I'd recommend making some `data class` to represent this with 3 fields--one for each tuple entry--and passing that in instead. This can go with the rest of the repo's data models, or just in this file if it's the only place you'll use it.

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+                        onGenderSelected: (String) -> Unit,
+                        onSportSelected: (String) -> Unit) {
+    var selectedOption by remember { mutableStateOf("All") }
```
thisjustin123, line 56: Formatting here is a bit weird. Plz make the first arg have its own line and do a reformat.

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+    var selectedSport by remember { mutableStateOf(sports.first().first) }
+
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(143.dp)
```
thisjustin123, line 62: Any need to explicitly set the height here?

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+        modifier = Modifier
+            .clickable(onClick = {
+                onSelect()
+            })
+            .width(111.dp)
+            .background(
+                color = if (selectedOption == option) CrimsonPrimary else Color.Transparent,
+                shape = RoundedCornerShape(100) // Rounded background
+            )
+            .padding(vertical = 8.dp, horizontal = 12.dp),
```
thisjustin123, line 121: Hmm, I probably would've preferred the background to slide behind the gender selectors and have it animate through. I guess this is ok for now.

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    SportSelectorHeader(sports, {}, {})
+}
+
+@Composable
+fun SportSelectorHeader(sports: List<Triple<String, Painter, Painter>>,
+                        onGenderSelected: (String) -> Unit,
+                        onSportSelected: (String) -> Unit) {
+    var selectedOption by remember { mutableStateOf("All") }
+    var selectedSport by remember { mutableStateOf(sports.first().first) }
```
thisjustin123, line 57: I'm a bit against having these as mutable states, since if we scroll away too far from the header for example, the state will be lost.

Let's take in the selectedOption/sport as an argument to the component, and expose a `onOptionSelected` and `onSportSelected`. This way, the viewmodel can handle the logic, which is more robust.

thisjustin123 (changes requested):
Really good start! Pretty clean code all things considered. Plz address comments though!

thisjustin123 (approved):
Looks epic 😎 thanks for incorporating changes!

PR #8, opened 2024-10-09
Implement empty state message

helenjb:
Screenshot 2024-10-09 at 17 50 20

- Implement UI for the empty state message
- Added/edited a few fonts and colors to align with design system


app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
```diff
@@ -0,0 +1,161 @@
...
+                .width(24.dp)
+                .height(24.dp)
+        )
+        Text(
+            text = option,
+            style = sportFilterText,
+            color = if (selectedOption == option) CrimsonPrimary else GrayMedium
+        )
+    }
+}
```
thisjustin123, line 161: Huh, this stuff shouldn't be in this PR right? This was in your other PR.

app/src/main/java/com/cornellappdev/score/components/EmptyStateMessage.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.cornellappdev.score.theme.Style.heading2
+import com.example.score.R
+
+@Preview
+@Composable
+fun PreviewEmptyStateMessage() {
+    Box(modifier = Modifier.background(Color.White)) {
+        EmptyStateMessage()
+    }
+}
```
thisjustin123, line 33: Plz put at bottom of file!

app/src/main/java/com/cornellappdev/score/components/EmptyStateMessage.kt
```diff
@@ -0,0 +1,59 @@
...
+
+@Composable
+fun EmptyStateMessage() {
+    Column(modifier = Modifier.width(158.dp),
```
thisjustin123, line 37: To ensure we can align this thing and whatnot in a screen later, let's take in a `modifier: Modifier = Modifier` and give it to the `Column` (i.e. `Column(modifier = modifier.width(158.dp),`)

thisjustin123: That way if we need to pass in some `Modifier.align()`, for example, we can do that

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
thisjustin123: I guess you can wait for the other to be merged, then rebase on main. The diff here should be cleaned up afterwards. Or, ideally, more work, but you can just manually pull out the changes here and put them in a new branch branched from `main`.

thisjustin123: To be clear, it seems like you made this branch off your other branch instead of `main`.

thisjustin123 (changes requested):
Weird git stuff, and a few small comments!

app/src/main/java/com/cornellappdev/score/components/SportSelectorHeader.kt
helenjb: I think this is just related to my Git/GitHub issues because I'm quite confident that I created this branch off of `main` :')

PR #9, opened 2024-10-23
UI Tweaks, LiveState

kelsiliul:
Added LiveState to the game card when a game is happening now;
image
 Changes in Home page UI.
image


app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -138,10 +149,26 @@ fun SportCard(
...
+                    )
+                } else {
+                    Row( horizontalArrangement = Arrangement.spacedBy(4.dp),
+                        verticalAlignment = Alignment.CenterVertically,) {
+                        Box(
+                            modifier = Modifier
+                                .padding(1.dp)
+                                .width(8.dp)
+                                .height(8.dp)
+                                .background(saturatedGreen, shape = CircleShape)
```
thisjustin123, line 167: We should make this opacity fade like a sine wave to communicate it being live. You should be able to do this with some sort of `animateFloatAsState()` ... `rememberInfiniteTransition()`... something like that. It's a very common problem, so I'd suggest finding an online / ChatGPT soln for this.

To reiterate, we should animate the opacity of this background as a sine wave. You can then do `saturatedGreen.copy(alpha = ...)` to input the alpha / opacity.

app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -47,16 +57,17 @@ fun SportCard(
         colors = CardDefaults.cardColors(containerColor = Color.White),
         modifier = modifier
             .clip(cardShape)
+            .shadow(elevation = 6.dp, spotColor = SpotColor, ambientColor = AmbientColor)
             .then(
                 if (topCornerRound) {
-                    Modifier.border(.4.dp, Color.Gray, cardShape)
+                    Modifier
+                        .border(width = 1.dp, color = Color(0xFFEAEAEA),cardShape)
```
thisjustin123, line 64: What's this color? I see it appearing a few times in this file.

If it's reused a lot and has a name, please pull this out into `Color.kt` as a new constant.

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -17,16 +17,19 @@ import com.cornellappdev.score.components.UpcomingGamesCarousel
...
+
 @Composable
 fun HomeScreen() {
     Column(verticalArrangement = Arrangement.spacedBy(16.dp, Alignment.Top))
     {
         UpcomingGamesCarousel(gameList)
-        Column(modifier = Modifier.padding(horizontal = 24.dp)) {
+
+        Column(verticalArrangement = Arrangement.spacedBy(16.dp, Alignment.Top),
+            modifier = Modifier.padding(horizontal = 24.dp)) {
```
thisjustin123, line 28: Nit: plz put these on different lines:

```kotlin
Column(
    verticalArrangement = Arrangement.spacedBy(16.dp, Alignment.Top),
    modifier = Modifier.padding(horizontal = 24.dp),
) {
// ...
```

thisjustin123 (changes requested):
One main improvement with animations!

thisjustin123 (approved):
Looks great :D

PR #15, opened 2025-02-05
Small UI Fixes

amjiao:
## Overview
- Miscellaneous ui fixes
## Changes Made
- Fixed the spacing of the upcoming games card so that it looks like one cohesive card
- Fixed spacing in the game score summary screen to center the column to the right of the icons
- Added custom colors with 40% transparency for the upcoming game card header
## Test Coverage
- see game score summary screen in preview
- everything else is visible in emulator 
## Next Steps
- colors are currently hard-coded in the UI, need to add a color field to data classes at some point or compute the color
## Screenshots
<details>
  <summary>color change</summary>
Screenshot 2025-02-05 at 2 25 26 PM
</details>
<details>
<summary>preview scoring details screen</summary>
Screenshot 2025-02-05 at 2 27 03 PM
</details>


app/src/main/java/com/cornellappdev/score/components/SportCard.kt
```diff
@@ -58,7 +58,7 @@ fun SportCard(
         colors = CardDefaults.cardColors(containerColor = Color.White),
         modifier = modifier
             .clip(cardShape)
-            .padding(8.dp)
+//            .padding(8.dp)
```
zachseidner1, line 61: Remove this rather than comment it out

app/src/main/java/com/cornellappdev/score/components/UpcomingGameCard.kt
```diff
@@ -145,7 +147,7 @@ fun GameScheduleScreen() {
         sportIcon = painterResource(id = R.drawable.ic_baseball),
         modifier = Modifier,
         headerModifier = Modifier,
-        gradientColor1 = Color(0xE1A69F),
-        gradientColor2 = Color(0x011F5B)
+        gradientColor1 = cornell_red,
+        gradientColor2 = penn_blue
```
zachseidner1, line 151: These should not be hardcoded, you need to extract the color from the school logo (this is specified in the [design](https://www.figma.com/design/eRiaEZ9MgOeCMEGf1xVAIV/Score-FA24?node-id=56-709&p=f&m=dev)).  However I don't think you should work on this until we're getting logos from the BE, since the image format may be different and I wouldn't want you to redo work. For now let's add a comment that this is a placeholder and we plan to extract colors from images once we get that from BE.

app/src/main/java/com/cornellappdev/score/components/UpcomingGamesCarousel.kt
```diff
@@ -86,8 +87,8 @@ fun UpcomingGamesCarousel(games: List<GameCardData>) {
                 location = game.location,
                 modifier = Modifier,
                 headerModifier = Modifier,
-                gradientColor1 = Color(0xFFE1A69F),
-                gradientColor2 = Color(0xFF011F5B)
+                gradientColor1 = cornell_red,//E1A69F),
+                gradientColor2 = penn_blue//FF011F5B)
```
zachseidner1, line 91: remove these comments after the colors

app/src/main/java/com/cornellappdev/score/theme/Color.kt
```diff
@@ -17,4 +17,9 @@ val PeachPrimary = Color(0xFFE1A69F)
...
 val saturatedGreen = Color(0xFF55B63C)
 val SpotColor = Color(0x12000000)
-val AmbientColor = Color(0x12000000)
\ No newline at end of file
+val AmbientColor = Color(0x12000000)
+
+//0x66 is 40% transparency
+val cornell_red = Color(0x66B31B1B)
+val penn_blue = Color(0x66021E59)
+val princeton_orange = Color(0x66FF6000)
```
zachseidner1, line 25: Follow the existing naming conventions when adding new colors, but I don't think we should even be adding these colors since we should be basing it off of the image. For now, just rename them, and leave a comment saying that they are placeholders and we plan to derive the colors from school logos. I would just take the average color, but could be worth asking how iOS does this.

app/src/main/java/com/cornellappdev/score/screen/GameScoreSummaryScreen.kt
```diff
@@ -33,47 +39,52 @@ fun ScoreEventItemDetailed(event: ScoreEvent) {
...
-                textAlign = TextAlign.Center
-            )
-
-            val (homeScore, awayScore) = event.score.split(" - ").map { it.toInt() }
-            Row(
-                horizontalArrangement = Arrangement.Center,
-                verticalAlignment = Alignment.CenterVertically,
-                modifier = Modifier.padding(top = 4.dp)
+            Column(
+                horizontalAlignment = Alignment.Start
```
zachseidner1, line 58: Generally nesting `Column` is something to avoid, because they can usually be collapsed into one. But here, the inner `Column` also makes this inaccurate to the design, since the score is left justified when it should be in the center.

app/src/main/java/com/cornellappdev/score/screen/GameScoreSummaryScreen.kt
```diff
@@ -33,47 +39,52 @@ fun ScoreEventItemDetailed(event: ScoreEvent) {
...
+                        text = " - ",
+                        style = bodyNormal,
+                        textAlign = TextAlign.Center
+                    )
+                    Text(
+                        text = awayScore.toString(),
+                        style = if (awayScore >= homeScore) bodyNormal.copy(fontWeight = FontWeight.Bold) else bodyNormal,
+                        textAlign = TextAlign.Center
+                    )
+                }
```
zachseidner1, line 87: According to the design, this entire thing should be a single center justified textbox! Design has given you an answer key on the best way to implement the UI, please use it.

app/src/main/java/com/cornellappdev/score/screen/GameScoreSummaryScreen.kt
```diff
@@ -33,47 +39,52 @@ fun ScoreEventItemDetailed(event: ScoreEvent) {
         modifier = Modifier
             .fillMaxWidth()
             .padding(vertical = 16.dp),
-        verticalAlignment = Alignment.Top
+        verticalAlignment = Alignment.Top,
+        horizontalArrangement = Arrangement.SpaceBetween
```
zachseidner1, line 43: The [design](https://www.figma.com/design/eRiaEZ9MgOeCMEGf1xVAIV/Score-FA24?node-id=1190-1591&m=dev) shows a margin of `16` between the play summary and the game information, and the a margin of `8` between the game information and school logo. Designers actually put a lot of time into figuring out the spacing of these things, and it's important we respect their decisions.

zachseidner1 (changes requested):
I appreciate your attempt at the small UI fixes Amy! Sorry if this may seem a bit harsh, but the component still does not match the design right now. I think we can definitely get there though! I want to show you some stuff with Figma during work session.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -41,7 +41,11 @@ data class ScoreEvent(
     val eventType: String,
     val score: String,
     val description: String? = null
-)
+) {
+    private val scoreTuple get() = score.split(" - ")
+    val homeScore get() = scoreTuple[0]
+    val awayScore get() = scoreTuple[1]
```
zachseidner1, line 47: This assumes the format of `score` always takes on `"xx - xx"`. This is probably an ok assumption, but we also don't want to crash if it's wrong. Instead, use `.getOrNull` to access elements in the array, and put a default value if it's not there (like `"?"`)

app/src/main/java/com/cornellappdev/score/screen/GameScoreSummaryScreen.kt
```diff
@@ -33,50 +44,51 @@ fun ScoreEventItemDetailed(event: ScoreEvent) {
...
+                    ) {
+                        append(event.homeScore)
+                    }
+                    withStyle(style = spanBodyNormal) {
+                        append(" - ")
+                    }
+                    withStyle(style = awayScoreStyle) {
+                        append(event.awayScore)
+                    }
+                }
```
zachseidner1, line 88: Beautiful, so much nicer than before!

app/src/main/java/com/cornellappdev/score/screen/GameScoreSummaryScreen.kt
```diff
@@ -86,6 +98,7 @@ fun ScoreEventItemDetailed(event: ScoreEvent) {
     }
 }
 
+
 @Preview(showBackground = true)
 @Composable
 fun PreviewScoringDetailsScreen() {
```
zachseidner1, line 104: nit: while we're here, mind making this preview function `private`?

zachseidner1 (approved):
Great job Amy! This is a huge improvement from before. Using the designs helps a lot lol. I appreciate you addressing everything.
BUT make sure to address my comment on the assumption about the form of the `score` text. Good to avoid any possible scenario when the app can crash.

PR #16, opened 2025-02-19
Update UI

helenjb:
## Changes Made
- Created NavigationHeader.kt component and added it to appropriate screens
- Edited GameScoreSummaryScreen.kt into a full screen


app/src/main/java/com/cornellappdev/score/components/NavigationHeader.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.cornellappdev.score.theme.AmbientColor
+import com.cornellappdev.score.theme.CrimsonPrimary
+import com.cornellappdev.score.theme.GrayLight
+import com.cornellappdev.score.theme.SpotColor
+import com.cornellappdev.score.theme.Style.bodyMedium
+import com.cornellappdev.score.theme.Style.heading2
+import com.cornellappdev.score.theme.White
+
+@Composable
+fun NavigationHeader(title: String) {
```
zachseidner1, line 39: You're gonna need to add an `onBackPressed` parameter to this component to deal with what happens if the user actually presses the back button.

app/src/main/java/com/cornellappdev/score/components/NavigationHeader.kt
```diff
@@ -0,0 +1,64 @@
...
+                         .height(24.dp),
+               )
+               Text(text = title, modifier = Modifier.align(Alignment.Center), style = heading2)
+          }
+     }
+}
+
+@Preview
+@Composable
+fun NavigationHeaderPreview() {
```
zachseidner1, line 62: Thanks for including a preview! Nit: Let's make it `private`

app/src/main/java/com/cornellappdev/score/components/NavigationHeader.kt
```diff
@@ -0,0 +1,64 @@
...
+fun NavigationHeader(title: String) {
+     Box(modifier = Modifier
+           .shadow(elevation=8.dp, clip = false, spotColor = Color.Black.copy(0.05f))
+          .background(Color.White)) {
+          Box(modifier = Modifier
+               .padding(start=24.dp, top=56.dp, bottom=12.dp, end=24.dp)
+               .background(Color.White)
+               .fillMaxWidth()
+               .height(27.dp)) {
+               Image(
```
zachseidner1, line 48: Use `Icon` instead of `Image` for SVG icons.

app/src/main/java/com/cornellappdev/score/components/NavigationHeader.kt
```diff
@@ -0,0 +1,64 @@
...
+                    painter = painterResource(id = R.drawable.ic_left_arrowhead),
+                    contentDescription = "Left Arrowhead",
```
zachseidner1, line 50: nit: `contentDescription` is for accessibility, so instead of describing the picture in this case, it could be better to describe the functionality, like "Back button", so if someone has this read to them through text to speech they know what it does and why they might want to press it.

zachseidner1 (commented):
This is some clean UI code, nice job! Just add the `onBackPressed` parameter for the `NavigationHeader`, and use `Icon` instead of `Image`, then you should be good to go!

zachseidner1 (approved):
Awesome job fixing my comments and the fonts!

PR #23, opened 2025-03-05
Add import

helenjb:
## Overview
Added import for fillMaxSize()


zachseidner1 (approved):
Please request my review next time. Also, we need to fix your emulator during the next Score work session.

Issue #21, opened 2025-03-03
`HomeViewModel` and `ScoreRepository` refactor


zachseidner1:
We merged #20 to unblock people, but I think there's still some improvements that could be made. 
## HomeViewModel improvements
- `HomeUiState` should be refactored to expose an `ApiResponse` instead of just a data class with data. 
  - This way, in the UI we can do a `when` statement, and show loading and error states appropriately. 
- Right now, we have a function `updateGameList` that we have to remember to call whenever we we receive an update from `scoreRepository.upcomingGamesFlow`. But the whole point of flows is that so we _don't_ have to remember to call these update functions and we can just have it update automatically!
  - So instead of making an update function, we should just use [`map`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/map.html), to transform the flow exposed by the repository into our `UiState` accordingly. If we do this, we can probably remove the `init` block entirely, since we'll just be sending down a mapped flow from the repository, so we don't need to attach any observers to it, we'll just exposing a read-only mapped version. 
- The date functions you made should go in a separate `DateUtil.kt` file. Also, you should try simplifying them using `DateTimeFormatter`
## ScoreRepository Improvements
Before asking for improvements for this one, I made a generic function `toResult` that looks like this:
```Kotlin
/**
 * Maps an ApolloResponse to a generic Kotlin result. It either provides the data with no errors, or
 * a failure response containing the error message in the throwable.
 */
fun <T : Operation.Data> ApolloResponse<T>.toResult(): Result<T> {
    if (hasErrors() || exception != null) {
        return Result.failure(
            exception?.cause ?: RuntimeException(
                errors?.firstOrNull()?.message ?: "Unknown error occurred"
            )
        )
    }
    return try {
        Result.success(dataOrThrow())
    } catch (e: NoDataException) {
        Result.failure(e)
    }
}
```
  I think we should use this function for all of our Apollo calls, this way we can more easily see if it has errors or if the response is clean. So add this function to a file, then call it in `ScoreRepository`. Then make sure to handle the error and loading states appropriately. We can just fetch this data once when the app loads, in the `init` block of `ScoreRepository`. Use `appScope` for your repository scope, which I provided you with.

PR #27, opened 2025-03-09
HomeViewModel/ScoreRepositoryRefactor

amjiao:
## Overview
Addressed issue #21 
UI looks the same


app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -35,24 +36,31 @@ class ScoreRepository @Inject constructor(
         _upcomingGamesFlow.value = ApiResponse.Loading
         try {
             val response = (apolloClient.query(GamesQuery()).execute())
-            val games = response.data?.games ?: emptyList()
-            Log.d("ScoreRepository", "response fetched successfully")
+            val result = response.toResult()
```
zachseidner1, line 39: nit: imo it would be nicer to condense these into 1 variable.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -35,24 +36,31 @@ class ScoreRepository @Inject constructor(
...
+                            teamName = game.team.name,
+                            teamColor = game.team.color,
+                            gender = game.gender,
+                            sport = game.sport,
+                            date = game.date,
+                            city = game.city
+                        )
+                    }
                 }
+                Log.d("ScoreRepository", "#games: ${list.size}")
```
zachseidner1, line 58: remove log

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -35,24 +36,31 @@ class ScoreRepository @Inject constructor(
...
-                        gender = game.gender,
-                        sport = game.sport,
-                        date = game.date,
-                        city = game.city
-                    )
+            if(result.isSuccess){
+                val games = response.data?.games ?: emptyList()
+                Log.d("ScoreRepository", "response fetched successfully")
+
+                val list: List<Game> = games.mapNotNull { game ->
```
zachseidner1, line 45: never have a variable called `list`, this tells me nothing about the variable.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -35,24 +36,31 @@ class ScoreRepository @Inject constructor(
...
+                            teamColor = game.team.color,
+                            gender = game.gender,
+                            sport = game.sport,
+                            date = game.date,
+                            city = game.city
+                        )
+                    }
                 }
+                Log.d("ScoreRepository", "#games: ${list.size}")
+                _upcomingGamesFlow.value = ApiResponse.Success(list.toList())
```
zachseidner1, line 59: Nit: No need to call `toList` here I don't think?

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -23,71 +23,197 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
+                                date = game.dateString,
+                                isLive = game.isLive,
+                                genderIcon = painterResource(game.genderIcon),
+                                sportIcon = painterResource(game.sportIcon),
+                                location = game.location,
+                                topCornerRound = true
+                            )
+                        }
+                    }
                 }
```
zachseidner1, line 98: Let's abstract this into a `GameContent` private Composable. It should take in the `uiState` as a parameter. This way the preview will actually work and we will be less deeply nested.

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -23,71 +23,197 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
+//        }
+//    }
+//}
+//
+//@Preview(showBackground = true)
+//@Preview(showBackground = true, uiMode = Configuration.UI_MODE_NIGHT_YES)
+//@Composable
+//private fun HomeScreenPreview() {
+//    HomeScreen()
+//}
```
zachseidner1, line 219: Never leave commented out code like this in a final PR.

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -0,0 +1,39 @@
+package com.cornellappdev.score.util
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Converts date of form String "month-abbr day (day-of-week)" (for example, "Apr 29 (Tue)") to a LocalDate object
+ */
+
+fun formatDate(strDate: String): LocalDate? {
```
zachseidner1, line 10: Let's call this `parseDate`, also document the behavior when the parsing fails.

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -0,0 +1,39 @@
...
+        LocalDate.parse("$subDate ${LocalDate.now().year}", formatter) //assumes current year
+    } catch (e: Exception) {
+        null
+    }
+}
+
+/**
+ * Converts from format "#xxxxxx" to a valid hex, with alpha = 40. Ready to be passed into Color()
+ */
+fun formatColor(color: String): Int {
```
zachseidner1, line 24: `formatColor` in the `DateUtil` file? 🤨 

also, we really shouldn't need this function at all. Can we refactor so `teamColor` is a `Color` instead of an `Int`?

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -0,0 +1,39 @@
...
+/**
+ *  Takes in a LocalDate? object and returns it as a String of format "MM/DD/YYY"
+ */
+fun dateToString(date: LocalDate?): String {
+    if (date == null) {
+        return "--"
+    }
+    //Log.d("HomeViewModel", "formatedDate: ${date.month.value}/${date.dayOfMonth}/${date.year}")
+    return "${date.month.value}/${date.dayOfMonth}/${date.year}"
+}
```
zachseidner1, line 39: Instead of having a function for this, we should just have a constant `DateTimeFormatter` initialized via `DateTimeFormatter.ofPattern`

app/src/main/java/com/cornellappdev/score/viewmodel/HomeViewModel.kt
```diff
@@ -1,51 +1,90 @@
...
 import java.time.LocalDate
 import javax.inject.Inject
 
 data class HomeUiState(
     val selectedGender: GenderDivision,
     val sportSelect: SportSelection,
     val selectionList: List<SportSelection>,
-    val upcomingGameList: List<GameCardData>,
-    // TODO Add remaining dynamic data for UI
+    val loadedState: ApiResponse<List<GameCardData>>
```
zachseidner1, line 23: Nit: `loadedState` doesn't tell me much about what this variable is.

app/src/main/java/com/cornellappdev/score/viewmodel/HomeViewModel.kt
```diff
@@ -1,51 +1,90 @@
...
-            (selectedGender == GenderDivision.ALL || game.gender == selectedGender.displayName)
-                    && (sportSelect is SportSelection.All || (sportSelect is SportSelection.SportSelect && game.sport == sportSelect.sport.displayName))
+        get() = when (loadedState) {
+            is ApiResponse.Success -> loadedState.data.filter { game ->
+                (selectedGender == GenderDivision.ALL || game.gender == selectedGender.displayName) &&
+                        (sportSelect is SportSelection.All || (sportSelect is SportSelection.SportSelect && game.sport == sportSelect.sport.displayName))
+            }
+
+            ApiResponse.Loading -> emptyList()
+            ApiResponse.Error -> emptyList()
```
zachseidner1, line 33: We shouldn't expose `filteredGames`, instead the loaded state should already just display the filtered games.
Uh oh, this just made me realize that we really should have a big refactor for all of this....
Basically, we want the filtered games to automatically update when a selected filter changes, so each filter should be a flow, and then we should combine them with the upcomingGamesFlow to get the UI state, so it always automatically has the latest value. I don't think this would recompose if you apply a filter as it is right now, I don't think, but not entirely sure.

app/src/main/java/com/cornellappdev/score/viewmodel/HomeViewModel.kt
```diff
@@ -1,51 +1,90 @@
...
+                                        sportIcon = Sport.fromDisplayName(game.sport)?.emptyIcon
+                                            ?: R.drawable.ic_empty_placeholder
+                                    )
+                                }
+                                .sortedBy { it.date }
+                            ApiResponse.Success(gameCards)
+                        }
+
+                        ApiResponse.Loading -> ApiResponse.Loading
+                        ApiResponse.Error -> ApiResponse.Error
```
zachseidner1, line 81: Mapping data in API responses like this is a common pattern we'll see a lot throughout the code. Pull from `main`, as I just added a function to map an API response based on applying a transform to its data as this does a lot of that for you.

Also, this is getting pretty deeply nested, so we should have an extension function on the `Game` data class that maps this to `GameCardData`

app/src/main/java/com/cornellappdev/score/viewmodel/HomeViewModel.kt
```diff
@@ -1,51 +1,90 @@
...
+
+            ApiResponse.Loading -> emptyList()
+            ApiResponse.Error -> emptyList()
         }
 }
 
 @HiltViewModel
 class HomeViewModel @Inject constructor(
-    private val rootNavigationRepository: RootNavigationRepository,
+    //private val rootNavigationViewModel: RootNavigationViewModel,
```
zachseidner1, line 39: don't leave commented out stuff here like this.

gradle/libs.versions.toml
```diff
@@ -12,6 +12,7 @@ activity = "1.8.0"
 constraintlayout = "2.1.4"
 runtimeAndroid = "1.7.2"
 apollo = "4.1.1"
+coreI18n = "1.0.0-alpha01"
```
zachseidner1, line 15: What's with these version changes, we should probably revert this file.

zachseidner1 (changes requested):
Amy, this is really great work, genuinely! I know I left a lot of comments, but it's so awesome to see that you have an understanding of MVVM. We're relatively close to this being ready, just address my comments and then you should be good.

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -23,71 +23,197 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
 
-@RequiresApi(Build.VERSION_CODES.O)//TODO - change the manifest or leave this?
 @Composable
 fun HomeScreen(
     homeViewModel: HomeViewModel = hiltViewModel()
 ) {
-    val uiState by homeViewModel.uiStateFlow.collectAsState()
+    val uiState = homeViewModel.collectUiStateValue()
+    val filteredGames = uiState.filteredGames
+    val firstThreeGames = filteredGames.subList(0, min(filteredGames.size, 3)) //TODO: is this okay to be in the view?
```
zachseidner1, line 37: Let's rename this `upcomingGames`, so this way it matches its section header. Also, we can pretty easily move this to the data class instead of deriving it in the view. Do it in the same way you did for `filteredGames`. Also, use `.take` instead of `.subList`.

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -23,71 +23,197 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
+                    SportSelectorHeader(
+                        sports = uiState.selectionList,
+                        selectedGender = uiState.selectedGender,
+                        selectedSport = uiState.sportSelect,
+                        onGenderSelected = { homeViewModel.onGenderSelected(it) },
+                        onSportSelected = { homeViewModel.onSportSelected(it) }
                     )
+                    LazyColumn(modifier = Modifier.padding(horizontal = 24.dp)) {
+                        items(filteredGames.size) { page ->
+                            val game = filteredGames[page]
```
zachseidner1, line 85: This is an anti-pattern, use the `items` function that takes in `List<T>` as a parameter instead so you can access the list elements directly as a parameter to the `items` lambda rather than using the access operator.

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -23,71 +23,197 @@ import androidx.hilt.navigation.compose.hiltViewModel
...
+                    Text(
+                        text = "Failed to load games. Please try again.",
+                    )
+                }
+            }
+
+            is ApiResponse.Success -> {
+                UpcomingGamesCarousel(firstThreeGames)
+                Column {
+                    Text(
```
zachseidner1, line 68: Please extract this into a `HomeContent` composable, so the preview for this screen will actually work.

zachseidner1 (commented):
Oops my old review told you to refactor all the filters to use flows, even though that's the way it _should_ be done, I realized it's probably not worth it for us right now. I want you to not have to go through an entire refactor so let's just leave a TODO comment for now. The current solution works and is more concise.

app/build.gradle.kts
```diff
@@ -71,6 +71,7 @@ dependencies {
     implementation("androidx.navigation:navigation-compose:2.8.2")
     implementation("androidx.compose.material3:material3:1.0.0")
     implementation("com.google.dagger:hilt-android:2.51.1")
+    implementation(libs.androidx.core.i18n)
```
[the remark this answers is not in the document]
amjiao: I can't remember anymore, I was having build problems for a bit so this might've been an attempted fix but everything seems to work fine without it so I'll delete it

amjiao:
I think the formatColor function needs to be kept since it parses the String into a valid Color input and applies the appropriate alpha, so I feel like it's a little long to put in one line and I also feel like it needs documentation? I just moved the function into the ScoreRepository which seems a bit off so let me know if you see a better way to approach this.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -1,25 +1,45 @@
...
+            dateString = parseDate(date)?.format(outputFormatter)
+                ?: date,
+            isLive = (LocalDate.now() == parseDate(date)),
+            location = city,
+            gender = gender,
+            genderIcon = if (gender == "Mens") R.drawable.ic_gender_men else R.drawable.ic_gender_women,
+            sport = sport,
+            sportIcon = Sport.fromDisplayName(sport)?.emptyIcon
+                ?: R.drawable.ic_empty_placeholder
+        )
```
zachseidner1, line 35: this is an interesting way of mapping `Game` to `GameCardData`, but it wasn't what I was going for. It doesn't make sense to have a different data class be a _property_ of the `Game` model. I was looking for an **extension function** on the `Game` model, something like:
```Kotlin
fun Game.toGameCardData(): GameCardData = TODO()
```

Even though this does work, it's unintuitive to have something like this as a property of a `Game`.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -34,28 +33,42 @@ class ScoreRepository @Inject constructor(
...
+}
+
+/**
+ * Converts from format "#xxxxxx" to a valid hex, with alpha = 40. Ready to be passed into Color()
+ */
+fun formatColor(color: String): Int {
+    val alpha = (40 * 255 / 100)// Convert percent to hex (0-255)
+    val colorInt = Integer.parseInt(color.removePrefix("#"), 16)
+    return (alpha shl 24) or colorInt
+}
```
zachseidner1, line 74: Oh I thought the `Color` constructor had a variant that took in a string. I guess this is fine, but I am a bit concerned that this is a bad abstraction. I think with a name like `formatColor` it would be easy for a future developer to miss the alpha value 40 and then be confused why their alpha value is lower than they expect, even with the documentation. I think it's ok if we want to have a method like `parseColor` that just converts a hex code string into a `Color` object. Then, after calling `parseColor`, you would juse use `.copy(alpha = 0.4f * 255)`, when you need the `alpha` value to be 40. I think this is less error prone, and still quite concise, since changing the alpha value is not hard work once we've parsed the color.

As for the location, I recommend we move this to another `Util` file, perhaps something like `ColorUtil`. 

Let me know if you agree with this!

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -0,0 +1,25 @@
+package com.cornellappdev.score.util
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Converts date of form String "month-abbr day (day-of-week)" (for example, "Apr 29 (Tue)") to a LocalDate object
+ * Returns null when parsing [strDate] fails
+ */
+fun parseDate(strDate: String): LocalDate? {
```
zachseidner1, line 10: Very optional nit: I think `parseDateOrNull` name goes along better with the Kotlin naming conventions (stuff like `firstOrNull`

app/src/main/java/com/cornellappdev/score/viewmodel/HomeViewModel.kt
```diff
@@ -1,51 +1,64 @@
...
+                copy(
+                    loadedState = response.map { list ->
+                        list.map { game ->
+                            game.toGameCardData
+                        }.filter { game ->
+                            game.date?.isAfter(LocalDate.now().minusDays(1)) ?: false
+                        }.sortedBy { it.date }
+                    }
+                )
+            }
```
zachseidner1, line 58: This is so much more readable, nice job!

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-agp = "8.8.2"
+agp = "8.7.3"
```
zachseidner1, line 2: why did you downgrade the `agp` version?

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -1,93 +1,141 @@
...
+            }
         }
     }
 }
 
-@Preview(showBackground = true)
-@Preview(showBackground = true, uiMode = android.content.res.Configuration.UI_MODE_NIGHT_YES)
+@Preview
 @Composable
 private fun HomeScreenPreview() {
```
zachseidner1, line 123: Thanks for getting this preview working! It helped me catch some design inconsistencies which I'm going to make issues about lol

zachseidner1 (commented):
Nice job with this refactor! We're almost there. I appreciate your work 🙏

app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
```diff
@@ -1,93 +1,141 @@
...
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = Color.White)
+    ) {
+        HomeContent(
+            HomeUiState(
+                selectedGender = GenderDivision.ALL,
+                sportSelect = SportSelection.All,
+                selectionList = emptyList(),
```
zachseidner1, line 133: wait also can we not have the selection  list be empty bc it's not representative of what the actual UI will look like.

gradle/libs.versions.toml
amjiao: I was messing with my ide version oops, I'll be more careful ab the files im committing in the future!

zachseidner1 (approved):
Amazing work! 🥳

PR #31, opened 2025-03-18
Implement navigation and past games screen UI/ViewModel/majority of networking

helenjb:
## Overview
Implemented Past Games (aka Scores) screen, with the associated ViewModel and most of networking. Also implemented bottom bar and navigation between game cards (on both Home and Past Games Screen) to a dummy game details screen.
## Changes Made
- Refactored some component names to make it more applicable to games that are not upcoming and/or to make the names be more descriptive of the component (e.g. `UpcomingGameCard` to `FeaturedGameCard`, `SportCard` to `GameCard`, `UpcomingGamesCarousel` to `GamesCarousel`)
- Implemented UI for `PastGameCard` and `PastGamesScreen` (`PastGameCard`s are properly spaced on this screen)
- Implemented `PastGamesViewModel`
- Implemented bottom bar that allows for navigation to between home screen and past games screen
- Implemented navigation from `GameCard`/`PastGameCard` to a dummy `GameDetailsPage` -- will currently navigate back to the correct page when clicking the back arrow on the `GameDetailsScreen`
- Implemented networking for `PastGamesScreen`
- Minor UI fix to `GameCard` so long event/university names don't push the sport/gender icon off the card
## Additional Notes
- The final scores in the past game cards were obtained by parsing a `result` string, which is oftentimes in the format Result, CornellScore-OpponentScore (e.g. "W, 2-1"). Not all of the strings are in this format (e.g. 4th of 6, 1498 points for women's Swimming and Diving), but in this case, the `cornellScore` and `otherScore` parameters of the game and associated card should be null, and as of right now, null-scored games are filtered out.
- Int and decimal scores should properly be displayed as such
- Filters do not work -- should be fixed by Amy's code


app/src/main/java/com/cornellappdev/score/components/GamesCarousel.kt
```diff
@@ -56,7 +55,7 @@ fun DotIndicator(
 }
 
 @Composable
-fun UpcomingGamesCarousel(games: List<GameCardData>) {
+fun GamesCarousel(games: List<GameCardData>, upcoming: Boolean) {
```
zachseidner1, line 58: I would prefer if we could have a `GamesCarouselVariant` enum instead of just a boolean. It's more extensible if we want to add more variants down the line. It's also more readable passing something like `UPCOMING_VARIANT` into the Composable rather than `true`. Or another approach to this refactor is to create some sort of `BaseGamesCarousel` and then you can pass arguments into it and create two components, an `UpcomingGamesCarousel` and `PastGamesCarousel` that both use it accordingly. It's up to you though.

app/src/main/java/com/cornellappdev/score/screen/PastGamesScreen.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.cornellappdev.score.components.SportSelectorHeader
+import com.cornellappdev.score.components.GamesCarousel
+import com.cornellappdev.score.components.PastGameCard
+import com.cornellappdev.score.model.GameCardData
+import com.cornellappdev.score.theme.Style.heading1
+import com.cornellappdev.score.theme.White
+import com.cornellappdev.score.viewmodel.HomeViewModel
+import com.cornellappdev.score.viewmodel.PastGamesViewModel
+
+@RequiresApi(Build.VERSION_CODES.O)//TODO - change the manifest or leave this?
```
zachseidner1, line 34: We should be able to get rid of this right?

app/src/main/java/com/cornellappdev/score/screen/PastGamesScreen.kt
```diff
@@ -0,0 +1,94 @@
...
+        verticalArrangement = Arrangement.spacedBy(16.dp, Alignment.Top),
+        modifier = Modifier.statusBarsPadding().background(White)
+    )
+    {
+        //TODO: check - displaying the earliest three games
+        GamesCarousel(
+            uiState.pastGameList.subList(
+                0,
+                minOf(3, uiState.pastGameList.size)
+            ), false
```
zachseidner1, line 52: The UI should not be responsible for this logic. Also this can be simplified with `.take`

app/src/main/java/com/cornellappdev/score/viewmodel/PastGamesViewModel.kt
```diff
@@ -0,0 +1,165 @@
...
+) {
+    val filteredGames: List<GameCardData>
+        get() = pastGameList.filter { game ->
+            (selectedGender == GenderDivision.ALL || game.gender == selectedGender.displayName)
+                    && (sportSelect is SportSelection.All || (sportSelect is SportSelection.SportSelect && game.sport == sportSelect.sport.displayName))
+        }
+}
+
+@HiltViewModel
+class PastGamesViewModel @Inject constructor(
```
zachseidner1, line 36: Can you merge Amy's PR and refactor this ViewModel so it has similar logic to her's?

app/src/main/java/com/cornellappdev/score/viewmodel/PastGamesViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+) : BaseViewModel<PastGamesUiState>(
+    PastGamesUiState(
+        selectedGender = GenderDivision.ALL,
+        sportSelect = SportSelection.All,
+        selectionList = Sport.getSportSelectionList(),
+        loadedState = ApiResponse.Loading
+    )
+) {
+    init {
+        scoreRepository.fetchGames()
```
zachseidner1, line 47: Let's remove this line. It will help make #35 easier, and we shouldn't be refetching the data every time past games is opened.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -15,8 +15,16 @@ data class Game(
...
     val date: String,
-    val city: String
-)
+    val city: String,
+    val cornellScore: Number? = null,
+    val otherScore: Number? = null,
+) {
+    val isPast: Boolean
+        get() = when {
+            parseDateOrNull(date)!! < LocalDate.now() -> true
```
zachseidner1, line 24: No non-null assertions. Instead just have a default value. We shouldn't crash the app if a game fails to parse.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -103,7 +114,12 @@ enum class GameStatus {
     COMPLETED
 }
 
-fun Game.toGameCardData(): GameCardData{
+enum class GamesCarouselVariant {
+    UPCOMING_VARIANT,
+    PAST_VARIANT
```
zachseidner1, line 119: nit: subscripting variant is redundant since the enum class already has the word Variant

app/src/main/java/com/cornellappdev/score/components/PastGameCard.kt
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+}
+
+@Preview(showBackground = true)
+@Composable
+private fun PastGameCardPreview() {
+    Column {
+        PastGameCard(
+            teamLogo = "https://cornellbigred.com/images/logos/penn_200x200.png?width=80&height=80&mode=max", //painterResource(id = R.drawable.penn_logo),
```
zachseidner1, line 208: nit: don't leave commented out code

app/src/main/java/com/cornellappdev/score/components/PastGameCard.kt
```diff
@@ -0,0 +1,219 @@
...
+    genderIcon: Painter,
+    sportIcon: Painter,
+    cornellScore: Number,
+    otherScore: Number,
+    modifier: Modifier = Modifier,
+    onClick: (Boolean) -> Unit = {}
+) {
+    val cornellWins = cornellScore.toFloat() > otherScore.toFloat()
+    val isHome = location == "Ithaca, NY"
+    val firstWins = (cornellWins && isHome) || (!cornellWins && !isHome)
```
zachseidner1, line 67: sorry if I missed this during my first review, but all of these shouldn't be derived in the view. I think it would be a lot easier if you just pass `Game` down to `PastGameCard`, then `firstWins` (which should have a clearer name because I feel like I'm not sure what that actually means) can just be derived in the data class.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -39,21 +39,26 @@ class ScoreRepository @Inject constructor(
             if (result.isSuccess) {
                 val games = result.getOrNull()
 
-                val upcomingGameslist: List<Game> =
+                val gamesList: List<Game> =
                     games?.games?.mapNotNull { game ->
+                        val scores = game?.result?.split(",")?.getOrNull(1)?.split("-")
+                        val cornellScore = scores?.getOrNull(0)?.toNumberOrNull()
+                        val otherScore = scores?.getOrNull(1)?.toNumberOrNull()
```
zachseidner1, line 46: It would be great if you can move what you wrote in the PR description to a code comment here for future developers to see.

zachseidner1 (commented):
Really great work, we're so close. I think if you just clean up the past games card a little bit, you should be good!

zachseidner1:
Also note that there are still some issues with the screen UI as a whole, which I'm not expecting you to get to in this PR. Specifically #36 , #32 , and #34 . But those can all be addressed later.

zachseidner1 (approved):
Thanks for resolving the second round of comments! Ship it 🚀 🔥

PR #33, opened 2025-03-22
Error and Loading State UI

amjiao:
## Overview
ErrorState, LoadingState, and GameDetailsLoadingState UI.
Added onRefresh function to HomeViewModel for TryAgain button functionality in ErrorState.
## Changes Made
ErrorState, LoadingState, GameDetailsLoadingState to be called wherever there are ApiResponse.Loading or ApiResponse.Error. 
ErrorState needs to be called within another Column with a Spacer underneath because it needs to sit on top of the bottomNavBar
LoadingState can take custom header strings, so it should be used in both HomeScreen and PastGamesScreen.
## Test Coverage
Already implemented in HomeScreen, works as expected. 
## Next Steps (delete if not applicable)
Should the ErrorState show up after a certain amount of time has passed in the LoadingState?
## Screenshots

<details>
  <summary>GameDetailsLoadingState</summary>
Screenshot 2025-03-22 at 3 07 04 PM
</details>
<details>
<summary>ErrorState</summary>
HomeError
</details>
<details>
<summary>LoadingState</summary>
HomeLoading
</details>


app/src/main/java/com/cornellappdev/score/components/LoadingState.kt
```diff
@@ -0,0 +1,126 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.cornellappdev.score.theme.GrayStroke
+import com.cornellappdev.score.theme.Style.heading1
+
+@Composable
+fun LoadingState(
```
zachseidner1, line 25: Nit: Maybe loading screen would be a better name, since this component is met as a full screen and has its layout as such

app/src/main/java/com/cornellappdev/score/components/LoadingStateBox.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.Dp
+import com.cornellappdev.score.theme.GrayStroke
+
+@Composable
+fun LoadingStateBox(cornerRoundness: Int, height: Dp) {
```
zachseidner1, line 14: Also I think it's better to just expose a Modifier parameter here instead of 2 separate parameters

zachseidner1 (commented):
Nice work Amy! There's some improvements to be made, and I'd love for you to implement the shimmer modifier extension function if you have time. But things look good!

app/src/main/java/com/cornellappdev/score/components/ErrorState.kt
```diff
@@ -0,0 +1,75 @@
...
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(422.dp),
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.SpaceBetween
+    ) {
+        Column(
+            horizontalAlignment = Alignment.CenterHorizontally
```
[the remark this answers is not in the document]
amjiao: i made a column for the feedback icon + text and nested that + the button within the main column, so i think both of these are necessary

app/src/main/java/com/cornellappdev/score/components/LoadingStateBox.kt
amjiao: i also want to have the default width be the maxWidth and I don't have a default value for the height so maybe it would be better to leave height as it's own parameter too?

app/src/main/java/com/cornellappdev/score/components/ErrorState.kt
```diff
@@ -0,0 +1,80 @@
...
+                    contentDescription = "refresh icon"
+                )
+                Text("Try again")
+            }
+        }
+        Spacer(modifier = Modifier.height(70.dp))
+    }
+}
+
+@Preview
```
zachseidner1, line 76: can we do 
`@Preview(showBackground = true)`

So we can see the component better through the preview?

app/src/main/java/com/cornellappdev/score/components/GameDetailsLoadingState.kt
```diff
@@ -0,0 +1,107 @@
...
+            Spacer(modifier = Modifier.height(16.dp))
+            LoadingStateBox(8, 48.dp)
+        }
+    }
+}
+
+@Preview
+@Composable
+private fun GameDetailsLoadingStatePreview() {
+    GameDetailsLoadingState()
```
zachseidner1, line 106: So because you used a Composition local provider, this breaks your previews since we didn't provide the value here. Could you make a custom preview function `ScorePreview` that is sort of like `ResellPreview`, and then provide the value through a custom `ScoreTheme` function? Sorry to have you just using the Resell code so much but I do think it's the best solution. I don't want you to just feel like you're copying and pasting code and not understanding why so please ask questions if you have them! Composition locals are a tricky topic.

app/src/main/java/com/cornellappdev/score/util/ColorUtil.kt
```diff
@@ -1,11 +1,61 @@
 package com.cornellappdev.score.util
 
+import androidx.compose.runtime.compositionLocalOf
 import androidx.compose.ui.graphics.Color
 
+val Wash = Color(0xFFF4F4F4)
+val Stroke = Color(0xFFD6D6D6)
+val LocalInfiniteLoading = compositionLocalOf<Float> { error("No infinite loading provided") }
```
zachseidner1, line 8: Move these to `Color.kt`

zachseidner1 (approved):
Please take a look at my comments, I'd like you to fix previews and establish a `ScorePreview` component for use down the line. I trust you to make these changes well so I am approving this now so you can merge after making those changes. Nice work!

app/src/main/java/com/cornellappdev/score/components/ErrorState.kt
```diff
@@ -73,7 +73,7 @@ fun ErrorState(
     }
 }
 
-@Preview
+@Preview(showBackground = true)
```
zachseidner1, line 76: Since you made `ScorePreview` take in `backgroundColor` as a parameter, can we migrate this to `= ScorePreview(backgroundColor = Color.White)` instead of using `@Preview(showBackground = true)`?

Sorry if I asked you to do this initially, but I think I meant that you should do this if you decided to not include that parameter on `ScorePreview`

app/src/main/java/com/cornellappdev/score/components/GameDetailsLoadingState.kt
```diff
@@ -102,6 +102,8 @@ fun GameDetailsLoadingState(
 
 @Preview
 @Composable
-private fun GameDetailsLoadingStatePreview() {
-    GameDetailsLoadingState()
+private fun GameDetailsLoadingStatePreview(){
+    ScorePreview {
```
zachseidner1, line 106: Better to use 
```Kotlin
private fun GameDetailsLoadingStatePreview() = ScorePreview {

}
```
To save an indent. This may seem trivial, but we're establishing a precedent that will be used for the rest of the lifespan of Score, so I think it's important here.

app/src/main/java/com/cornellappdev/score/components/LoadingScreen.kt
```diff
@@ -123,11 +123,15 @@ private fun LoadingFilter() {
...
+    ScorePreview {
+        LoadingFilter()
+    }
 }
 
 @Preview
 @Composable
 private fun LoadingScreenPreview() {
-    LoadingScreen("Loading Upcoming...", "Loading Schedules...")
+    ScorePreview {
```
zachseidner1, line 134: Migrate to `= ScorePreview` like above

app/src/main/java/com/cornellappdev/score/components/LoadingStateBox.kt
```diff
@@ -40,5 +40,7 @@ fun LoadingStateBox(
 @Preview
 @Composable
 private fun LoadingStateBoxPreview() {
-    LoadingStateBox(12, 16.dp)
+    ScorePreview {
```
zachseidner1, line 43: Migrate to `= ScorePreview` like above

app/src/main/java/com/cornellappdev/score/components/ScorePreview.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.compose.ui.unit.dp
+import com.cornellappdev.score.theme.LocalInfiniteLoading
+
+@Composable
+fun ScorePreview(
+    padding: Dp = 0.dp,
+    backgroundColor: Color = Color.White,
+    content: @Composable () -> Unit
+) {
+    CompositionLocalProvider(
```
zachseidner1, line 20: You can provide the value the same way that you provided it before:
```Kotlin
val transition = rememberInfiniteTransition()

    val animatedValue = transition.animateFloat(
        initialValue = 0f,
        targetValue = 1f,
        animationSpec = InfiniteRepeatableSpec(
            animation = keyframes {
                durationMillis = 2000
                0f at 0
                1f at 1000
                0f at 2000
            }
        ),
        label = "infinite loading"
    ).value

    CompositionLocalProvider(
        LocalInfiniteLoading provides animatedValue
    ) {
```
Let me know if this makes sense!

app/src/main/java/com/cornellappdev/score/theme/Color.kt
```diff
@@ -22,4 +23,8 @@ val AmbientColor = Color(0x12000000)
 //placeholders, will be replaced once we get backend data
 val CornellRed = Color(0x66B31B1B)
 val PennBlue = Color(0x66021E59)
-val PrincetonOrange = Color(0x66FF6000)
\ No newline at end of file
+val PrincetonOrange = Color(0x66FF6000)
+
+val Wash = Color(0xFFF4F4F4)
+val Stroke = Color(0xFFD6D6D6)
+val LocalInfiniteLoading = compositionLocalOf<Float> { error("No infinite loading provided") }
```
zachseidner1, line 30: My bad for missing this during my first review, but I think that the `LocalInfiniteLoading` variable specifically should be in a separate file, it doesn't make sense in `Color.kt`. We can create a new file `CompositionLocals` in `theme` for composition locals that are related to app theme / style.

zachseidner1 (changes requested):
The animation makes this look so much better! There are some comments I left for you to address. Then I'll do a final pass over and we should be good.

zachseidner1 (approved):
This looks great, thanks for addressing my other comments! I did a quick migration of the rest of the previews to use `ScorePreview` so it's obvious to future developers that they need to use this. Please merge it when you're ready 🥳

PR #37, opened 2025-03-24
Emil jiang gamedetails UI

EmilJiang:
So, for gamedetails there are still a lot of things to work on. I will attach a picture of what it looks like now, and a list of things to work on. I just wanted to create this PR for comments because there is a lot of changes, and I am not totally done with everything. I tried to follow the code structure of Amy's as much as possible :). 

**What I did**

The networking should be all done, but sometimes it does not load which I think is a backend issue. In the slack, Skye said he will look into it. The Screen with all the components is mostly done. 

**Specific things that still needs work**

hopfully quick ones
- Timer for games that have not started yet 
- Add to calendar
- No scoring summary state
Screenshot 2025-03-23 at 9 54 26 PM

- Scoring Summary navigation (not sure if helen already added this)

longer ones
- Score for the Header (right now i have 0's as placeholders). I do not know how IOS is getting there scores from. A potential idea is summing up the periods from the scoreBreakdown, BUT sometimes that is null, so maybe continuously checking if result is not null signaling the end of match
- scoreBreakdown for each sport. There is a pdf for how the breakdown works, and I am thinking just to do a huge when statement of if statement. Right now it is a basic score of last one in the list being final score and the rest being the periods
- change scoreSummary to be what IOS has instead of figma. Left is Figma, right is IOS.
Screenshot 2025-03-23 at 9 50 53 PM
Screenshot 2025-03-23 at 9 51 58 PM

**Current result**
Screenshot 2025-03-23 at 9 53 34 PM


zachseidner1 (changes requested):
Pass CI before requesting review please. Thank you!

app/src/main/java/com/cornellappdev/score/viewmodel/GameDetailsViewModel.kt
```diff
@@ -1,20 +1,39 @@
...
+            loadedState = ApiResponse.Loading
     )
 ) {
-    data class GameDetailsUiState(
-        val homeScore: Int,
-        val awayScore: Int,
-    )
+    init {
+        val gameId: String? = savedStateHandle["gameId"]
+        scoreRepository.getGameById(gameId ?: "")
```
zachseidner1, line 28: I think instead of sending empty string to the backend, I think we should only send the request if `gameId` is non-null. I would want you to use `.let` for this, and I think the structure is a little unintuitive, so here's what I'm imagining:
```Kotlin
val gameId: String? = savedStateHandle["gameId"]
        gameId?.let {
            scoreRepository.getGameById(it)
            asyncCollect(scoreRepository.currentGamesFlow) { response ->
                applyMutation {
                    copy(
                        loadedState = response.map { gameCard ->
                            gameCard.toGameCardData()
                        }
                    )
                }
            }
        } ?: applyMutation {
            copy(loadedState = ApiResponse.Error)
        }
```
Lmk if that looks good to you.

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
```diff
@@ -0,0 +1,47 @@
+package com.cornellappdev.score.util
+
+import com.cornellappdev.score.model.GameData
+import com.cornellappdev.score.model.TeamBoxScore
+import com.cornellappdev.score.model.TeamScore
+
+fun toGameData(
```
zachseidner1, line 7: Can you write a method spec explaining how this function works, also you should link to the PDF file that explains how to parse the scores that I assume you referenced when making this?

This is optional given our time budget, but it would be amazing if you could write unit tests for this function lol.

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +26,30 @@ fun parseDateOrNull(strDate: String): LocalDate? {
  */
 val outputFormatter = DateTimeFormatter.ofPattern("M/d/yyyy")
 
+fun parseDateTimeOrNull(strDate: String, strTime: String): LocalDateTime? {
```
zachseidner1, line 29: Write a method spec for this, it should explain what the date and time strings could look like.

zachseidner1, line 29: Nit: `str` should not be in the variable name as it's already in the type.

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +26,30 @@ fun parseDateOrNull(strDate: String): LocalDate? {
...
+
+    return try {
+        LocalDateTime.parse(dateTimeString, formatter)
+    } catch (e: Exception) {
+        null
+    }
+}
+
+
+fun formatDateTimeDisplay(strDate: String, strTime: String): String {
```
zachseidner1, line 48: Again we need a method spec here. Also same nit with having `str` in the name.

app/src/main/java/com/cornellappdev/score/components/ScoreSummary.kt
```diff
@@ -47,13 +49,25 @@ fun ScoreEventItem(event: ScoreEvent) {
...
         verticalAlignment = Alignment.CenterVertically
     ) {
-        Image(
-            painter = painterResource(event.team.logo),
-            contentDescription = event.team.name,
-            modifier = Modifier
-                .size(40.dp)
-                .padding(end = 12.dp)
-        )
+        if (event.team.name == "COR"){ // TODO: Check if its "COR" for all queries. It is for baseball
```
zachseidner1, line 52: I made #38 so we don't forget about this.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -63,4 +67,30 @@ class ScoreRepository @Inject constructor(
             _upcomingGamesFlow.value = ApiResponse.Error
         }
     }
+
+    /**
+     * Asynchronously fetches game details for a particular game. Once finished, will send down
+     * `currentGamesFlow` to be observed.
```
zachseidner1, line 73: Nit: This function doesn't really "send down" anything, it more just updates the current games flow

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -63,4 +67,30 @@ class ScoreRepository @Inject constructor(
...
+                val game = result.getOrNull()
+                if (game?.game != null) {
+                    _currentGameFlow.value = ApiResponse.Success(game.game.toGameDetails())
+                } else {
+                    Log.e("ScoreRepository", "Game or game.game is null for id: $id")
+                    _currentGameFlow.value = ApiResponse.Error
+                }
+            } else {
+                _currentGameFlow.value = ApiResponse.Error
+            }
```
zachseidner1, line 89: This whole block can be simplified:
```Kotlin
result.getOrNull()?.game?.let {
                _currentGameFlow.value = ApiResponse.Success(it.toGameDetails())
            } ?: _currentGameFlow.update { ApiResponse.Error }
 ```
This is another reason why we use result. `getOrNull` returns non-null data if and only if the result is a success, so it's very concise to work with.

File: app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt [hunk 8]
```diff
@@ -1,58 +1,130 @@
...
+                fontSize = 18.sp,
+                modifier = Modifier
+                    .weight(1f)
+            )
+            Text(
+                text = "Game Details",
+                fontSize = 18.sp,
+                textAlign = TextAlign.Center,
+                modifier = Modifier.weight(2f)
+            )
```
zachseidner1, line 106: Use the `NavigationHeader` component instead, and make it outside of the `GameDetailsContent` so this way error and loading states still have the header shown.

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -67,123 +139,44 @@ fun GameDetailsScreen() {
...
-
-//     GameDetailsScreen(
-//         scoreEvents = scoreEvents,
-//         onArrowClick = {
-//             println("Arrow clicked to view more details")
-//         }
-//     )
-}
\ No newline at end of file
+private fun GameDetailsScreenPreview() {}
```
zachseidner1, line 182: No empty previews

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -67,123 +139,44 @@ fun GameDetailsScreen() {
...
-                Text(text = "9/28/2024, 2:00PM", style = bodyNormal.copy(color = GrayPrimary))
+                Text(text = gameCard.dateString, style = bodyNormal.copy(color = GrayPrimary))
             }
 
             //render the below if the game is in the future
-            Spacer(modifier = Modifier.height(40.dp))
-            TimeUntilStartCard(2, 0)
-
+            if(gameCard.isPastStartTime){
+                if(!gameCard.scoreBreakdown.isNullOrEmpty()){
```
zachseidner1, line 158: Nit: I think `gameCard.scoreBreakdown?.isNotEmpty() == true` is more readable

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -1,58 +1,130 @@
...
+import com.cornellappdev.score.viewmodel.HomeUiState
+import com.cornellappdev.score.viewmodel.HomeViewModel
 
 @Composable
-fun GameDetailsScreen() {
-    Column(modifier = Modifier.background(White).fillMaxSize(), horizontalAlignment = Alignment.CenterHorizontally) {
-        // TODO: add navigation
-        NavigationHeader(title = "Game Details", {})
+fun GameDetailsScreen(
+    gameId: String,
```
zachseidner1, line 51: remove this unused parameter

zachseidner1 (changes requested):
Emil, this is a really strong first draft, especially as a new member. Really great work. There are still some fixes to be made, but you did a nice job. Also please provide a way for me to test the game details screen! Right now the function is just unused.

zachseidner1 (commented):
image <br />
I found some more UI issues: you should update this screen so that the padding matches the [Figma page](https://www.figma.com/design/FKyvYncmFJhitMnB1Yurve/Score-SP25?node-id=1274-2333&m=dev). Please use dev mode to see the padding easier, and add the missing score header.

We can fix the score box card in a separate PR since that's not something you worked on.

EmilJiang:
UI fixes still need to be done, but I think all the logic should be done.

EmilJiang:
I have added all the functionality, and done the UI changes for most. I am still confused on the navigation header padding for the top. In the image below, it cant be 56 because that includes the top of the phone too I think. 
Screenshot 2025-04-09 at 1 03 03 AM
Testing is the same as before with a button on the top which sometimes does not work if scraping is done. Some things that still need to be done are navigation and scoring summary UI fix to match with IOS. I will ask for the design tomorrow
Screenshot 2025-04-09 at 1 05 51 AM

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +27,58 @@ fun parseDateOrNull(strDate: String): LocalDate? {
...
 
+/**
+ * Parses a date and time string into a [LocalDateTime] object.
+ *
+ * Expected input format:
+ * - [date]: String of form "MMM d (EEE)". For example, "Apr 5 (Sat)"
+ * - [time]: String of form "h:mm a" with or without periods. For example, "3:45 PM" or "3:45 p.m."
+ *
+ * The current year is assumed in the final parsed result.
+ * returns [LocalDateTime] object if parsing succeeds, or [null] if the format is invalid.
```
zachseidner1, line 38: Nit: can you fix this to be a javadoc style comment? With `@param` and `@returns`

Otherwise this is a really nice spec!

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +27,58 @@ fun parseDateOrNull(strDate: String): LocalDate? {
...
+
+
+/**
+ * Formats a date and time string into a user-friendly display string.
+ *
+ * Combines [date] and [time] inputs and attempts to parse them using [parseDateTimeOrNull].
+ * If parsing succeeds, returns a string in the format "MMM d, h:mma" (For example, "Apr 5, 3:45PM").
+ * If parsing fails, returns an empty string.
+ */
+fun formatDateTimeDisplay(date: String, time: String): String {
```
zachseidner1, line 66: Can you document the parameters here too like you did above?

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +27,58 @@ fun parseDateOrNull(strDate: String): LocalDate? {
...
+    val dateTime = parseDateTimeOrNull(date, time)
+    val formatter = DateTimeFormatter.ofPattern("MMM d, h:mma", Locale.ENGLISH)
+
+    return dateTime?.format(formatter) ?: ""
+}
+
+fun getTimeUntilStart(date: String, time: String): Pair<Int, Int>? {
```
zachseidner1, line 73: This needs a spec

app/src/main/java/com/cornellappdev/score/util/DateUtil.kt
```diff
@@ -23,3 +27,58 @@ fun parseDateOrNull(strDate: String): LocalDate? {
...
+    val gameStart = parseDateTimeOrNull(date, time) ?: return null
+    val now = LocalDateTime.now()
+
+    if (gameStart.isBefore(now)) return null
+
+    val duration = Duration.between(now, gameStart)
+    val days = duration.toDays().toInt()
+    val hours = duration.minusDays(days.toLong()).toHours().toInt()
+    return Pair(days, hours)
```
zachseidner1, line 82: Nit: `days to hours` reads slightly better imo

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
```diff
@@ -0,0 +1,71 @@
+package com.cornellappdev.score.util
+
+import com.cornellappdev.score.model.GameData
+import com.cornellappdev.score.model.TeamBoxScore
+import com.cornellappdev.score.model.TeamScore
+
+// TODO: ASK ABOUT OT 
+fun convertScores(scoreList: List<String?>?, sport: String): Pair<List<Int>, Int?> {
```
zachseidner1, line 8: This needs a spec

zachseidner1, line 8: Could you provide the PDF that BE provided so I can double check the accuracy of this function?

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
```diff
@@ -0,0 +1,71 @@
...
+        scoresByPeriod = scoresByPeriod.take(9)
+        val totalScore = scoresByPeriod.sum()
+        return Pair(scoresByPeriod, totalScore)
+    }
+
+    val totalScore = scoreList.last()?.toIntOrNull()
+    return Pair(scoresByPeriod, totalScore)
+}
+
+fun toGameData(
```
zachseidner1, line 31: This needs a spec

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    val team1Score =
+        TeamScore(team = team1, scoresByPeriod = team1Scores, totalScore = team1Total ?: 0)
+    val team2Score =
+        TeamScore(team = team2, scoresByPeriod = team2Scores, totalScore = team2Total ?: 0)
+
+    return GameData(teamScores = Pair(team1Score, team2Score))
+}
+
+fun parseResultScore(result: String?): Pair<Int, Int>? {
```
zachseidner1, line 53: This needs a spec

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
```diff
@@ -0,0 +1,71 @@
...
+    team2: TeamBoxScore,
+    sport: String
+): GameData {
+    val (team1Scores, team1Total) = if (!scoreBreakdown.isNullOrEmpty())
+        convertScores(scoreBreakdown[0], sport)
+    else Pair(emptyList(), null)
+
+    val (team2Scores, team2Total) = if (scoreBreakdown != null && scoreBreakdown.size > 1)
+        convertScores(scoreBreakdown[1], sport)
+    else Pair(emptyList(), null)
```
zachseidner1, line 43: Here is a more Kotlin way of doing the same thing:
```Kotlin
val (team1Scores, team1Total) = scoreBreakdown?.getOrNull(0)?.let {
        convertScores(it, sport)
    } ?: (emptyList<Int>() to null)

    val (team2Scores, team2Total) = scoreBreakdown?.getOrNull(1)?.let {
        convertScores(it, sport)
    } ?: (emptyList<Int>() to null)
```

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -10,52 +11,113 @@ import androidx.compose.foundation.layout.fillMaxSize
...
 @Composable
-fun GameDetailsScreen(gameId: String = "", onBackArrow: () -> Unit = {}) {
-    Column(modifier = Modifier.background(White).fillMaxSize(), horizontalAlignment = Alignment.CenterHorizontally) {
-        // TODO: add navigation
-        NavigationHeader(title = "Game Details", onBackArrow)
+fun GameDetailsScreen(
+    gameDetailsViewModel: GameDetailsViewModel = hiltViewModel(),
+    onBackArrow: () -> Unit = {}
+) {
+    val uiState by gameDetailsViewModel.uiStateFlow.collectAsState()
```
zachseidner1, line 51: Can be simplified to:
```Kotlin
val uiState = gameDetailsViewModel.collectUiStateValue()
```

zachseidner1 (changes requested):
Great job on this Emil! So happy to see this page coming together. I left a few comments for you to address 🫡

app/src/main/java/com/cornellappdev/score/components/NavigationHeader.kt
```diff
@@ -1,68 +1,59 @@
...
-                              .width(24.dp)
-                              .height(24.dp),
-                    )
-               }
-               Text(text = title, modifier = Modifier.align(Alignment.Center), style = heading2, color = Color.Black)
-          }
-     }
+    Box(
+        modifier = Modifier
+//           .shadow(elevation=8.dp, clip = false, spotColor = Color.Black.copy(0.05f)) todo: what is this for
```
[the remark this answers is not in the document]
EmilJiang: Here is the header from figma.
Screenshot 2025-04-09 at 5 57 19 PM
I thought that the shadow was kind of weird
Screenshot 2025-04-09 at 5 57 51 PM

app/src/main/java/com/cornellappdev/score/util/GameDataUtil.kt
EmilJiang: [score-breakdown.pdf](https://github.com/user-attachments/files/19686966/score-breakdown.pdf)

zachseidner1 (changes requested):
hey sorry for responding to this a bit late, but thank you for pushing those fixes! for the boolean, navigateToGameDetails should not take in a boolean as an argument. It should take in a string instead. Then we should pass in the string to the navController here [instead ](https://github.com/cuappdev/score-android/blob/b51b13351e1b95a6243cfcf7d7bed6951752161a/app/src/main/java/com/cornellappdev/score/nav/root/RootNavigation.kt#L108-L110)of the empty string. Also, do you think you’re able to resolve merge conflicts on your own, or do you want me to help out with that? it seems like there are a lot so lmk what you think.

EmilJiang:
I added navigation to the components to gameDetails. One thing I noticed though was the padding is kind of off. There seems to be hardcoded values, but figma does not show. The "add to calendar" is off the page. 
Screenshot 2025-04-14 at 5 42 09 PM
There no numbers attached to spacing between days left and add to calendar. Also, I do not see an example with the bottom navigation. So, I do not know what the padding should be with the bottom navigation or if it should be scrollable
Screenshot 2025-04-14 at 5 43 46 PM

zachseidner1 (commented):
For situations like these, we can use Spacers with `weight` modifier. This way it will take up as much space as it can, but it will never cut off the button. We can kinda eye-ball the Figma as a general guideline for weight ratios.

app/src/main/java/com/cornellappdev/score/components/FeaturedGameCard.kt
```diff
@@ -123,6 +124,7 @@ private fun FeaturedGameCardPreview() = ScorePreview {
 
 @Composable
 fun FeaturedGameCard(
+    id: String,
```
zachseidner1, line 127: We should never require an `id` to show a Composable, the only parameters Composables should take are things related to its view state, or click methods. So really, this composable should have an `onClick: () -> Unit` method. Then we would write:
```Kotlin
onClick = { 
                    onClick(game.id)
                }
```

And pass it through like that. Lmk if that makes sense.

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -161,28 +162,35 @@ fun GameDetailsContent(gameCard: DetailsCardData) {
...
+                    }
+
+                    Spacer(modifier = Modifier.weight(1f))
+
+                    ButtonPrimary(
+                        "Add to Calendar",
+                        painterResource(R.drawable.ic_calendar),
+                        onClick = {
+                            gameCard.toCalendarEvent()?.let { event ->
+                                addToCalendar(context = context, event)
```
zachseidner1, line 194: Optional nit: I don't think you need to pass in context as a named parameter, feels a bit redundant.

app/src/main/java/com/cornellappdev/score/components/PastGameCard.kt
```diff
@@ -47,7 +47,7 @@ import java.time.LocalDate
 fun PastGameCard(
     data: GameCardData,
     modifier: Modifier = Modifier,
-    onClick: (Boolean) -> Unit = {}
+    onClick: (String) -> Unit = {}
```
zachseidner1, line 50: Same thing here. Ideally this should just be of type `() -> Unit`, then the caller would handle the ID. The `GameCardData` is a view state, and we shouldn't need to pass in an ID to render a preview of it.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -90,6 +91,7 @@ class ScoreRepository @Inject constructor(
     fun getGameById(id: String) = appScope.launch {
         _currentGameFlow.value = ApiResponse.Loading
         try {
+            Log.d("ScoreRepository", "getGameById: ${id}")
```
zachseidner1, line 94: Remove logs

app/src/main/java/com/cornellappdev/score/util/TestingConstants.kt
```diff
@@ -28,6 +29,7 @@ val PENN_GAME = GameCardData(
 )
 
 val PRINCETON_GAME = GameCardData(
+    id = "1",
```
zachseidner1, line 32: See we don't want to have to add these useless IDs to our test data for our previews.

zachseidner1 (commented):
Thank you for responding to my comments! I have a few more 😭  but these are mostly minor. I want to say you did a really great job with this! I really appreciate you being receptive and being able to address most things I say quite quickly. 

Make sure to pull before beginning work again since I made a couple small fixes to this branch.

Also, the scoring summary shouldn't be a `LazyColumn`, it should just only show the first 3 items. You can use Kotlin's `Iterable<T>.take` extension function for this.

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
EmilJiang: The addTocalendar still does not work, so I think I will be working on that as the first issue.

EmilJiang:
I think in this case, GameCardData needs to have an id because the uistate uses GameCardData. If there is a better way, please let me know!

zachseidner1:
> I think in this case, GameCardData needs to have an id because the uistate uses GameCardData. If there is a better way, please let me know!

Are you sure add to calendar doesn't work? I tried it and it opened google calendar, but not much else happened. Does the same issue happen on your end?

zachseidner1 (approved):
Finally ready to merge this in! Thank you so much again for responding to all of my comments. I think we've had good discussion and I hope you've learned a lot!

Yes that was my bad about the `GameCardData`, it's fine for it to have an ID. But it definitely shouldn't be a parameter to the Composable and I'm glad you were able to remove that.

Issue #44, opened 2025-04-09
Fix scoring box summary UI


zachseidner1:
It doesn't match the design.

PR #48, opened 2025-04-14
Fix ScoreBox UI

amjiao:
## Overview
Address issue #44 
## Changes Made
Rounded corners and included border
## Test Coverage
Checked visually, matches Figma


app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -35,54 +38,54 @@ fun BoxScore(gameData: GameData) {
...
     )
-
-    Column(
+    Surface(
         modifier = Modifier
-            .fillMaxWidth()
-            .background(Color.White, shape = RoundedCornerShape(8.dp))
+            .fillMaxWidth(),
+        shape = RoundedCornerShape(8.dp),
+        border = BorderStroke(width = 1.dp, color = CrimsonPrimary)
```
zachseidner1, line 46: I don't think you need a `Surface` here. These should be able to just go as modifiers on the `Column`

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -35,54 +39,54 @@ fun BoxScore(gameData: GameData) {
...
-                modifier = Modifier.weight(1f),
-                style = bodyNormal,
-                color = Color.White,
-                textAlign = TextAlign.Center
+            TeamScoreRow(
+                teamScore = gameData.teamScores.first,
+                totalTextColor = saturatedGreen,
             )
-        }
+            Divider(color = CrimsonPrimary, thickness = 1.dp)
```
zachseidner1, line 83: Divider is deprecated, replace with `HorizontalDivider`

zachseidner1 (changes requested):
Thank you for working on this Amy. I think there are some code quality improvements you could make for the UI. Additionally, the main thing this was supposed to fix is how long scores appear, which your change does not address. See how the preview I added displays differently than the Figma? I think you'll need a minimum width on the text, and also account for text overflow and use ellipses

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
zachseidner1: Oh sorry ignore this for now, I was reviewing with a later version of material3. I'm gonna bump material3 on `main`, then you can merge main into your branch, invalidate caches and restart, then you should be able to see `HorizontalDivider`

zachseidner1: you should be good to change this to `HorizontalDivider` now!

amjiao:
- Edited GameDetailsScreen lines because the ScoreBox covers the case where the score breakdown is empty (with the placeholder hyphens) and ScoreBox should always display if current time is past the game start time.
- Didn't implement the team icon instead of team name change because I think I'd have to mess around with data classes and I should prob focus on finishing other issues first

zachseidner1 (approved):
This is just barely ready for MVP but there are still some issues that I'd like you to resolve.
image
We're missing the drop shadow around the score box. The border gets cut off at the corner. Also, the school names should have a min width of 60 as specified by the design. Also, the numbers for the periods are not center aligned with their scores.

PR #50, opened 2025-04-14
Limit sport filters for MVP

### LINKED ISSUE #39: Should only display select sports for MVP
Display:

- soccer
- football
- ice hockey
- field hockey
- lacrosse
- baseball

And basketball, after backend is fixed

helenjb:
- Fixes #39 , #49, #53

### DIFF
--- a/app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
+++ b/app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
@@ -12,6 +12,7 @@ import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import javax.inject.Inject
 import javax.inject.Singleton
+import com.cornellappdev.score.util.isValidSport
 
 /**
  * This is a singleton responsible for fetching and caching all data for Score.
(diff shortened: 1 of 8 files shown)


zachseidner1 (changes requested):
Thank you for the fix! Instead of rewriting this if statement so many times, can we factor this out into a function that checks if a sport display name is valid (something that we want to show)? Typing out so many strings again and again like this is extremely error prone.

zachseidner1 (changes requested):
Could you resolve the merge conflicts before requesting my review? On Android Studio, pull the latest changes from main, then go back to your branch and do Git > Merge > origin/main. Then this should bring up the merge conflict resolution editor. Let me know if you need help from there!

helenjb:
For some reason when I run the app with these changes, on the first run it will fail to fetch games (or it times out), but when I rerequest on the same run, it will successfully load. Not sure if this is an issue with my emulator or the code.

app/src/main/java/com/cornellappdev/score/model/Sport.kt
```diff
@@ -173,7 +174,7 @@ enum class Sport(
                 GenderDivision.FEMALE -> Sport.entries.filter { it.gender == GenderDivision.FEMALE || it.gender == GenderDivision.ALL }
                 GenderDivision.ALL,
                 null -> Sport.entries
-            }
+            }.filter { isValidSport(it.displayName) }
```
zachseidner1, line 177: I don't think we need to filter here since we already should've done this in the repository.

app/src/main/java/com/cornellappdev/score/model/ScoreRepository.kt
```diff
@@ -45,34 +46,35 @@ class ScoreRepository @Inject constructor(
...
-                                sport = game.sport,
-                                date = game.date,
-                                city = game.city,
-                                cornellScore = cornellScore,
-                                otherScore = otherScore
-                            )
-                        }
-                    } ?: emptyList()
+                    games?.games?.filter { game -> isValidSport(game?.sport ?: "") }
+                        ?.mapNotNull { game ->
```
zachseidner1, line 50: Nit: `filterNotNull` and then `map` is a bit cleaner here so you don't have to handle `null` in the predicate.

zachseidner1 (approved):
Thanks for making these changes! I think this is the usual behavior of Score because of the backend, so don't worry about the loading issue. It could also be an emulator thing.

app/src/main/java/com/cornellappdev/score/model/Sport.kt
helenjb: This does not work if I remove it

zachseidner1: oh oops that was definitely wrong of me, I thought that was data coming from the BE, sorry for not looking at that closer.

PR #72, opened 2025-04-18
Edit error state to accommodate game details screen

helenjb:
Addressed #59


zachseidner1 (approved):
Change seems reasonable, thanks for doing this!

Issue #61, opened 2025-04-16
Scoring Box takes a minimum of 4 scores


EmilJiang:
There are games with less than 4 scores, and this should be displayed correctly.

Issue #60, opened 2025-04-16
Disable Add to Calendar


EmilJiang:
Currently, clicking the "Add to Calendar" on GameDetails Screen only opens the calendar. No event is added.

PR #73, opened 2025-04-21
Emil jiang navigate

EmilJiang:
I added navigation in the game details screen. I am not sure if this is the best way to do it, but it was the only way I could think of after trying many things. To test, the best way I found was to go into past scores and go from there as many of the future games do not have scoring summaries. #46


app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -164,10 +170,26 @@ fun GameDetailsContent(gameCard: DetailsCardData) {
...
+                        )
+                        Spacer(modifier = Modifier.weight(1f))
+                        IconButton(onClick = { navigateToGameScoreSummary(gameCard.scoreEvent) }) {
+                            Icon(
+                                painter = painterResource(id = R.drawable.ic_right_chevron),
+                                contentDescription = "Back button",
+                                modifier = Modifier
+                                    .width(24.dp)
+                                    .height(24.dp),
+                            )
```
zachseidner1, line 189: image
the icon appears really off here?

zachseidner1 (changes requested):
After a bit of research I found out that this _should_ definitely be possible to do this with type safety. See [this Stack Overflow](https://stackoverflow.com/q/79298677), we need to use the `typeMap` parameter of `composable` in our root navigator. If you have time today, I'd love for you to take a crack at it. Otherwise we can leave it like this and create a GitHub issue, because it's not that bad.

Regardless of what you decide to do, you need to fix the icon for navigating to the scoring summary.

EmilJiang:
I saw that stackoverflow, but I did not know how to add the typeMap as that requires some sort of external library, and I think I need to create a ListNavType which I do not really know how to do. 
The arrow is fixed. 
Screenshot 2025-04-21 at 9 19 17 PM
By disable in #60, do you just mean remove? If so, my new commit address it. 
Fixed the ScoreBox #61. This has not been thoroughly tested as the backend is really slow for me.
Image is now correct with the team for Scoring Summary, and ScoringSummaryScreen. Checked the real players name to make sure it matched the team.
Screenshot 2025-04-21 at 9 56 52 PM
Screenshot 2025-04-21 at 9 57 47 PM

zachseidner1 (approved):
LGTM 🚀

PR #83, opened 2025-09-21
ScoreBox fix (+ lacrosse icon mini fix)

### LINKED ISSUE #80: Scoring summary UI additional fixes
Fallout from #48 

image <br />
We're missing the drop shadow around the score box. The border gets cut off at the corner. Also, the school names should have a min width of 60 as specified by the design. Also, the numbers for the periods are not center aligned with their scores.

amjiao:
## Overview
Fixes #80 and #66 
Also fixes small bug with the logic of coloring the winning team's total score green.

## Changes Made
- Fixed lacrosse icon
- Fixed border cutoffs by wrapping ScoreBox component in a Surface
- Fixed team name cutoff error by setting a max width of 70.dp (approximately how long "Cornell" is in the font size for > 4 data columns) to allow adequate space for both the team name column and all of the data columns (if there's < 4 data columns, then I allow the team name column to take up IntrinsicSize.Max to fill up empty space
- Fixed alignment issues by refactoring table to a Row of Columns structure as opposed to a Column of Rows.
- Added if-else statement in the total scores column so that the color is green if that team won
- Accounted for edge case where the number of data columns > 10, in which case the data columns become a LazyRow that scrolls between the fixed team name column and totals column

## Test Coverage
- Wrote more test data for different game lengths (2, 4, 6, 10, 12 in total)
- appears correct in previews

Screenshot 2025-09-21 at 7 16 27 PM

### DIFF
--- a/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
+++ b/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
@@ -110,8 +110,7 @@ private fun TeamNameColumn(
             //set to approximation of minimum width for "Cornell"
             modifier.widthIn(max = 70.dp)
         } else {
-            //can take up as much space as is available
-            modifier.width(IntrinsicSize.Max)
+            modifier.widthIn(max = 100.dp)
         }
     ) {
         Row(
@@ -294,7 +293,7 @@ private fun CompleteTableData(
             teamTwoScores = gameData.teamScores.second,
             //if maxPeriods > 8, "Totals" header will wrap to two lines. In this case, don't weight so that space is allocated to TotalsColumn first
             //otherwise, TotalsColumn will fit without wrapping and can be allocated equal width as the data columns
-            modifier = if (maxPeriods < 8) {
+            modifier = if (maxPeriods < 4) {
                 Modifier.weight(1f, fill = true)
             } else {
                 Modifier
@@ -386,38 +385,39 @@ private fun TotalsColumn(
 }
 
 
+//Padding added to previews to simulate the padding around the ScoreBox when it's displayed on the screen
 @Preview
 @Composable
 private fun PreviewBoxScore() = ScorePreview {
-    BoxScore(gameData = gameData)
+    BoxScore(gameData = gameData, Modifier.padding(start = 20.dp, end = 20.dp))
 }
 
 @Preview
 @Composable
 private fun PreviewBoxScoreForLongGame() = ScorePreview {
-    BoxScore(longGameData)
+    BoxScore(longGameData, Modifier.padding(start = 20.dp, end = 20.dp))
 }
 
 @Preview
 @Composable
 private fun PreviewBoxScoreForMedGame() = ScorePreview {
-    BoxScore(mediumGameData)
+    BoxScore(mediumGameData, Modifier.padding(start = 20.dp, end = 20.dp))
 }
 
 @Preview
 @Composable
 private fun PreviewBoxScoreForShortGame() = ScorePreview {
-    BoxScore(shortGameData)
+    BoxScore(shortGameData, Modifier.padding(start = 20.dp, end = 20.dp))
 }
 
 @Preview
 @Composable
 private fun PreviewBoxScoreExtraLongGame() = ScorePreview {
-    BoxScore(gameData = extraLongGameData)
+    BoxScore(gameData = extraLongGameData, Modifier.padding(start = 20.dp, end = 20.dp))
 }
 
 @Preview
 @Composable
 private fun PreviewBoxScoreEmpty() = ScorePreview {
-    BoxScore(gameData = emptyGameData())
+    BoxScore(gameData = emptyGameData(), Modifier.padding(start = 20.dp, end = 20.dp))
 }
--- a/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
+++ b/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
@@ -47,8 +47,8 @@ import com.cornellappdev.score.util.mediumGameData
 import com.cornellappdev.score.util.shortGameData
 
 private val HEADER_HEIGHT = 35.dp
-private val SMALL_BOX_SIZE = 4
-private val LARGE_BOX_SIZE = 10
+private const val SMALL_BOX_SIZE = 4
+private const val LARGE_BOX_SIZE = 10
 
 @Composable
 fun BoxScore(
--- a/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
+++ b/app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
@@ -27,6 +27,8 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.tooling.preview.PreviewParameterProvider
 import androidx.compose.ui.unit.dp
 import com.cornellappdev.score.model.GameData
 import com.cornellappdev.score.model.ScoresByPeriod
@@ -78,10 +80,18 @@ fun BoxScore(
                 rowTextStyle,
                 gameData.maxPeriods
             )
-            CompleteTableData(
-                gameData = gameData,
-                rowTextStyle = rowTextStyle
-            )
+            if (gameData.maxPeriods > LARGE_BOX_SIZE) {
+                CompleteLazyTableData(
+                    gameData = gameData,
+                    rowTextStyle = rowTextStyle
+                )
+            } else {
+                CompleteTableData(
+                    gameData = gameData,
+                    rowTextStyle = rowTextStyle,
+                    maxPeriods = gameData.maxPeriods
+                )
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+private fun PreviewBoxScoreExtraLongGame() = ScorePreview {
+    BoxScore(gameData = extraLongGameData, Modifier.padding(start = 20.dp, end = 20.dp))
+}
 
 @Preview
 @Composable
-private fun PreviewTeamScoreRow() = ScorePreview {
-    TeamScoreRow(gameData.teamScores.first, GrayMedium, 4, bodyNormal)
+private fun PreviewBoxScoreEmpty() = ScorePreview {
+    BoxScore(gameData = emptyGameData(), Modifier.padding(start = 20.dp, end = 20.dp))
```
zachseidner1, line 422: mega nit: A fun thing you could do here to improve the code is to use parametrized previews so you don't need to write 5 preview functions (remember Art of The Preview tech talk 😁 ). But up to you if you find value in taking time to do this.

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
         gameData.teamScores.first.scoresByPeriod.size,
         gameData.teamScores.second.scoresByPeriod.size
     )
-    val rowTextStyle = if (maxPeriods > 4) labelsNormal else bodyNormal
+
+    val rowTextStyle = if (maxPeriods > 4) {
+        metricSmallNormal
+    } else {
+        metricNormal
+    }
```
zachseidner1, line 62: Nit: you could declare this variable closer to where it is used to make the code more readable

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
 @Composable
-fun BoxScore(gameData: GameData) {
+fun BoxScore(
+    gameData: GameData,
+    modifier: Modifier = Modifier
+) {
     val maxPeriods = maxOf(
         gameData.teamScores.first.scoresByPeriod.size,
         gameData.teamScores.second.scoresByPeriod.size
     )
```
zachseidner1, line 56: I wonder why we need this variable, I suppose we can't assume that these values will be the same? This could be good to extract to a derived property of `GameData`.

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
-    val rowTextStyle = if (maxPeriods > 4) labelsNormal else bodyNormal
+
+    val rowTextStyle = if (maxPeriods > 4) {
```
zachseidner1, line 58: we should extract 4 to a constant like you did for `HEADER_HEIGHT`

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+            modifier = Modifier
+                .fillMaxWidth()
+        ) {
+            TeamNameColumn(
+                gameData.teamScores.first.team.name,
+                gameData.teamScores.second.team.name,
+                rowTextStyle,
+                maxPeriods
+            )
+            if (maxPeriods > 10) {
```
zachseidner1, line 84: same thing with 10, we should extract it to a well-named constant

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
-        modifier = Modifier
-            .fillMaxWidth()
-            .clip(shape = RoundedCornerShape(8.dp))
-            .background(color = Color.White, shape = RoundedCornerShape(8.dp))
-            .border(BorderStroke(width = 1.dp, color = CrimsonPrimary))
+        modifier = if (maxPeriods > 4) {
+            //set to approximation of minimum width for "Cornell"
+            modifier.widthIn(max = 70.dp)
+        } else {
+            modifier.widthIn(max = 100.dp)
```
zachseidner1, line 113: Just curious did you try my recommendation of using the auto sizing text?

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+        }
+    ) {
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .background(color = CrimsonPrimary)
+                .height(HEADER_HEIGHT)
+                .padding(top = 6.dp, bottom = 4.dp),
+        ) {}
```
zachseidner1, line 122: Could we just use `Spacer` here instead?

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+                color = GrayPrimary,
+                modifier = Modifier
+                    .weight(1f, fill = true)
+                    .padding(10.dp),
+                textAlign = TextAlign.Left,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis
+            )
+        }
+        HorizontalDivider(thickness = 1.dp, color = CrimsonPrimary)
```
zachseidner1, line 143: The design doesn't have any horizontal dividers, maybe we should just use `Spacer` here if you need this for the arrangement or eliminate this?

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis
+            )
+        }
+    }
+}
+
+@Composable
+private fun TableDataColumn(
+    header: Int,
```
zachseidner1, line 169: nit: I think it makes more sense for this component to receive `String` for the header, since it is just text. Might help make the UI a bit more readable

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
-            maxPeriods,
-            rowTextStyle
-        )
+@Composable
+private fun CompleteLazyTableData(
+    gameData: GameData,
+    rowTextStyle: TextStyle,
+    modifier: Modifier = Modifier
+) {
+    val periodScores = mapToPeriodScores(gameData)
```
zachseidner1, line 229: Ideally `mapToPeriodScores` should be done in the VM, and the `ScoreBox` component as a whole should accept `ScoreBoxViewState` instead of `GameData`. But if you don't want to refactor all of this right now we don't have to. For now, you could just make `toPeriodScores` a function on the `GameData` `data class`.

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+
+    if (periodScores.isNotEmpty()) {
```
zachseidner1, line 231: Nit: could do `if (periodScores.isEmpty()) return` to reduce nesting

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+    modifier: Modifier = Modifier
+) {
+    val periodScores = mapToPeriodScores(gameData)
+
+    Row(
+        modifier = modifier,
+        horizontalArrangement = Arrangement.SpaceEvenly
+    ) {
+        if (periodScores.isNotEmpty()) {
+            mapToPeriodScores(gameData).map { periodScore ->
```
zachseidner1, line 271: nit: could just use `periodScores` directly instead of calling the `mapToPeriodScores` function again

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+                    teamTwoScore = "-",
+                    rowTextStyle = rowTextStyle,
+                    modifier = Modifier.weight(1f)
+                )
+            }
+        }
+        TotalsColumn(
+            teamOneScores = gameData.teamScores.first,
+            teamTwoScores = gameData.teamScores.second,
+            //if maxPeriods > 8, "Totals" header will wrap to two lines. In this case, don't weight so that space is allocated to TotalsColumn first
```
zachseidner1, line 294: you mean >= 4 in this comment? That is what your code seems to suggest

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
```diff
@@ -24,164 +29,395 @@ import androidx.compose.ui.tooling.preview.Preview
...
+    val periodScores = mapToPeriodScores(gameData)
+
+    if (periodScores.isNotEmpty()) {
+        Row(
+            modifier = modifier
+        ) {
+            LazyRow(
+                modifier = Modifier.weight(1f)
+            ) {
+                items(periodScores) { periodScore ->
```
zachseidner1, line 238: The only thing changing here is the period scores row. I wonder if we could just have the lazy and non lazy versions as separate composables, instead of having the entire table datas as separate composables? Lmk if that makes sense

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
amjiao: prob just a preview rendering issue since the structure is a row of columns.

app/src/main/java/com/cornellappdev/score/components/ScoreBox.kt
amjiao: the headers for this column will always be ints because it's a period number, is it still better to have headers be strings?

PR #85, opened 2025-10-01
Advnced Filter UI

EmilJiang:
## Overview
Added advanced filter button and buttonsheet. 

## Test Coverage
Visual Tests

## Next Steps 
ViewModel Logic
Related filtering logic: [Related filtering logic](https://github.com/cuappdev/eatery-blue-android/blob/9d9b764eda88a0d3dd8b1f4119d8e54dc8911db4/app/src/main/java/com/cornellappdev/android/eatery/ui/viewmodels/CompareMenusBotViewModel.kt#L67)

Screenshot 2025-10-01 at 5 25 13 PM

Screenshot 2025-10-01 at 5 25 50 PM

### DIFF
--- a/app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
+++ b/app/src/main/java/com/cornellappdev/score/screen/HomeScreen.kt
@@ -29,6 +29,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
@@ -36,6 +37,7 @@ import androidx.hilt.navigation.compose.hiltViewModel
 import com.cornellappdev.score.components.ButtonPrimary
 import com.cornellappdev.score.components.EmptyStateBox
 import com.cornellappdev.score.components.ErrorState
+import com.cornellappdev.score.components.ExpandableSection
 import com.cornellappdev.score.components.GameCard
 import com.cornellappdev.score.components.GamesCarousel
 import com.cornellappdev.score.components.LoadingScreen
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/cornellappdev/score/components/ButtonPrimary.kt
```diff
@@ -32,7 +39,9 @@ fun ButtonPrimary(text: String, icon: Painter?, onClick: () -> Unit = {}) {
                     .height(24.dp),
                 colorFilter = ColorFilter.tint(White)
             )
-            Spacer(modifier = Modifier.width(8.dp))
+            if (text.isNotEmpty()) {
+                Spacer(modifier = Modifier.width(8.dp))
+            }
         }
         Text(text = text, style = bodyMedium.copy(color = White))
```
zachseidner1, line 46: nit: imo it makes more sense to both wrap the `Text` and the `Spacer` in `text.isNotEmpty()` rather than just the `Spacer`

app/src/main/java/com/cornellappdev/score/components/ExpandableSection.kt
```diff
@@ -0,0 +1,66 @@
...
+            Icon(
+                painter = painterResource(
+                    id = if (expanded) R.drawable.ic_round_minus else R.drawable.ic_round_plus
+                ),
+                contentDescription = if (expanded) "Collapse" else "Expand"
+            )
+        }
+
+        if (expanded) {
+            Column {
```
zachseidner1, line 47: Have you tried having this conditional rendering within the `Column`, and then given the `Column` the `animateContentSize` modifier? I think it would make this look much better!

app/src/main/java/com/cornellappdev/score/components/ExpandableSection.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    Column {
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .clickable { expanded = !expanded },
+            horizontalArrangement = Arrangement.SpaceBetween,
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            Text(title, fontSize = 18.sp)
```
zachseidner1, line 37: The font for this is wrong. You should be using the `h2` text style for this according to the Figma. If this isn't in the code base yet, you need to add it. 

Image

All this typography should be implemented as specified by the Figma.

app/src/main/java/com/cornellappdev/score/components/ExpandableSection.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.cornellappdev.score.R
+
+@Composable
+fun ExpandableSection(title: String, options: List<String>) {
+    var expanded by remember { mutableStateOf(false) }
+    var selectedOption by remember { mutableStateOf("Under $20") }
```
zachseidner1, line 27: This non-nullable type is asserting a filter is always selected. What about when there are no filters selected? Also, I think we should be using an `enum class` with a display string property to store the filters, rather than just strings.

app/src/main/java/com/cornellappdev/score/components/ExpandableSection.kt
```diff
@@ -0,0 +1,66 @@
...
+            Column {
+                options.forEach { option ->
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .clickable { selectedOption = option }
+                            .padding(vertical = 4.dp)
+                    ) {
+                        RadioButton(
```
zachseidner1, line 56: You are just using the default material `RadioButton`. As a result the colors are wrong

app/src/main/java/com/cornellappdev/score/components/ExpandableSection.kt
```diff
@@ -0,0 +1,66 @@
...
+                            selected = (selectedOption == option),
+                            onClick = { selectedOption = option }
+                        )
+                        Text(option)
```
zachseidner1, line 60: Text is styled incorrectly

app/src/google-services.json
zachseidner1, line 1: This file should not be committed. We should get rid of it and do the proper set up using GitHub secrets. See `hustle-android` for a recent example of this.

zachseidner1 (changes requested):
Please double check that your UI matches the Figma for future PRs before requesting my review. But thank you for your work!

PR #86, opened 2025-10-24
Highlights Screen UI

amjiao:
## Overview
Highlights screen UI
## Changes Made
Highlights Card UI + main highlights screen
## Test Coverage
Previews visually match the figma
## Next Steps (delete if not applicable)
- Make the highlights search screen
- Include highlights button in bottom nav bar
Screenshot 2025-10-24 at 1 20 51 PM


app/src/main/java/com/cornellappdev/score/components/EmptyState.kt
```diff
@@ -24,6 +24,7 @@ import com.cornellappdev.score.theme.Style.heading2
 @Composable
 fun EmptyState(
     modifier: Modifier = Modifier,
+    icon: Int = R.drawable.ic_speaker_gray,
```
AndrewCheung360: Nit: could use @DrawableRes for the icon to ensure passed in int matches a valid resource

AndrewCheung360: Also, not necessarily a thing to remove, but is there a reason this has a default argument?

app/src/main/java/com/cornellappdev/score/components/EmptyState.kt
```diff
@@ -24,6 +24,7 @@ import com.cornellappdev.score.theme.Style.heading2
...
     title: String = "No games yet.",
```
AndrewCheung360: Is there a reason for default arg?

app/src/main/java/com/cornellappdev/score/components/EmptyState.kt
```diff
@@ -24,6 +24,7 @@ import com.cornellappdev.score.theme.Style.heading2
...
     subtitle: String = "Check back here later!"
```
AndrewCheung360: Same as above

app/src/main/java/com/cornellappdev/score/components/EmptyState.kt
```diff
@@ -53,6 +54,7 @@ fun EmptyState(
 fun EmptyStateBox(
     modifier: Modifier = Modifier,
     height: Dp = 550.dp,
+    icon: Int = R.drawable.ic_speaker_gray,
     title: String = "No games yet.",
```
AndrewCheung360: same as above

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+import com.cornellappdev.score.theme.CrimsonPrimary
+import com.cornellappdev.score.theme.GrayStroke
+import com.cornellappdev.score.theme.Style.bodySemibold
+import com.cornellappdev.score.theme.Style.heading2
+import com.cornellappdev.score.theme.Style.labelsNormal
+import com.cornellappdev.score.theme.White
+
+@Composable
+fun VideoHighlightCardHeader(
+    image: String
```
AndrewCheung360: Nit: I think imageUrl would be slightly better

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+) {
+    Box(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(117.dp)
+            .clip(RoundedCornerShape(topStart = 16.dp, topEnd = 16.dp))
```
AndrewCheung360: I think it should be possible to apply the rounded corner clip modifier to the top level column that contains this header

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+    videoHighlight: VideoHighlightData,
+    wide: Boolean,
+    modifier: Modifier = Modifier
+) {
+    Column(
+        modifier = if (wide) modifier.fillMaxWidth() else modifier.width(241.dp)
+    ) {
+        VideoHighlightCardHeader(videoHighlight.image)
+
+        Column(
```
AndrewCheung360: Nit: It might be more readable to extract this as a separate component

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+            Spacer(Modifier.height(8.dp))
+            Row(
+                modifier = Modifier.fillMaxWidth(),
+                horizontalArrangement = Arrangement.SpaceBetween,
+                verticalAlignment = Alignment.CenterVertically
+            ) {
+                Row(
+                    verticalAlignment = Alignment.CenterVertically
+                ) {
+                    BasicText(
```
AndrewCheung360: I think it would be good to add a TODO or github issue to handle sending an intent to open up the Youtube app with that url on click

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+                    style = labelsNormal,
+                    text = "11/09"
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun ArticleHighlightCard(
```
AndrewCheung360: Nit: I think there is a decent amount of code in this file, so it could be worth considering separating article highlight card from video highlight card

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+                    color = Color.White,
+                    style = labelsNormal,
+                    text = "11/09"
+                )
+            }
+        }
+    }
+}
+
+@Preview
```
AndrewCheung360: Nit: I think it would be good to use the preview parameter providers instead of making multiple previews for the same component

app/src/main/java/com/cornellappdev/score/components/HighlightsFilter.kt
```diff
@@ -0,0 +1,89 @@
...
+        modifier = Modifier
+            .height(32.dp)
+            .border(
+                width = 1.dp,
+                color = GrayStroke,
+                shape = RoundedCornerShape(100)
+            )
+            .background(color = Color.White, shape = RoundedCornerShape(100))
+            .padding(horizontal = 12.dp, vertical = 4.dp)
+            .clip(shape = RoundedCornerShape(100))
```
AndrewCheung360: is this needed with the background rounded corner shape?

app/src/main/java/com/cornellappdev/score/components/HighlightsFilter.kt
```diff
@@ -0,0 +1,89 @@
...
+    LazyRow(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(start = 24.dp),
+        horizontalArrangement = Arrangement.SpaceBetween,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        items(sportList.size) { i ->
+            HighlightsFilterButton(sportList[i], { onFilterSelected(sportList[i]) })
+            Spacer(Modifier.width(12.dp))
```
AndrewCheung360: I think you can just use Arrangement.spacedBy in the horizontalArrangement argument instead of using spacer

app/src/main/java/com/cornellappdev/score/components/HighlightsFilter.kt
```diff
@@ -0,0 +1,89 @@
...
+    onFilterSelected: (Sport) -> Unit,
+) {
+    LazyRow(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(start = 24.dp),
+        horizontalArrangement = Arrangement.SpaceBetween,
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        items(sportList.size) { i ->
```
AndrewCheung360: I see that sportlist is in testing constants. We might want to consider moving that out of testing constants since it is used in an actual component not just for testing purposes. Also, not sure if those 3 sports are the only ones supported right now, but if not, we might want to update that list as well

Nit: you can also use the other items syntax without needing to use [i]

app/src/main/java/com/cornellappdev/score/model/Highlights.kt
```diff
@@ -0,0 +1,17 @@
...
+    val image: String,
+    val url: String,
+    val date: String,
+    val sport: Sport,
+    val gender: GenderDivision
+)
+
+data class ArticleHighlightData(
+    val title: String,
+    val image: String,
```
AndrewCheung360: same as above

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+import com.cornellappdev.score.theme.GrayLight
+import com.cornellappdev.score.theme.GrayMedium
+import com.cornellappdev.score.theme.Style.bodyNormal
+import com.cornellappdev.score.theme.Style.heading1
+import com.cornellappdev.score.theme.Style.heading2
+import com.cornellappdev.score.util.highlightsList
+import com.cornellappdev.score.util.sportList
+
+@Composable
+fun HighlightsScreenHeader(
```
AndrewCheung360: If we choose to keep it in this file, this should be private unless it will be used elsewhere

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+        )
+        Spacer(modifier = Modifier.height(12.dp))
+        Row(
+            modifier = Modifier
+                .padding(horizontal = 24.dp)
+                .border(width = 1.dp, color = GrayLight, shape = RoundedCornerShape(100.dp))
+                .background(color = Color.White, shape = RoundedCornerShape(100.dp))
+                .fillMaxWidth()
+                .clip(shape = RoundedCornerShape(100.dp))
+                .clickable { onSearch },
```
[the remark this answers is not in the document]
AndrewCheung360: On that note, we should probably have some text that will be updated when the user types, but I think we can add a todo since that will likely be linked to the viewmodel uistate once that is added

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+    onSearch: () -> Unit
+) {
+    Column(modifier = Modifier.fillMaxWidth()) {
+        Text(
+            modifier = Modifier.padding(start = 24.dp),
+            style = heading1,
+            text = "Highlights"
+        )
+        Spacer(modifier = Modifier.height(12.dp))
+        Row(
```
AndrewCheung360: Nit: Could be worth extracting as a separate component for the search bar

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        Spacer(modifier = Modifier.height(16.dp))
+        HighlightsFilterRow({/*todo on filter selected*/})
+    }
+}
+
+@Composable
+fun HighlightsCardRow(
+    highlightsList: List<Any>
```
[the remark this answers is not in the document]
AndrewCheung360: Maybe you can use a sealed class or something like that to give the two different data types are shared data type of HighlightData

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+}
+
+@Composable
+@Preview
+private fun HighlightsScreenPreview() {
+    HighlightsScreen(highlightsList, highlightsList)
+}
+
+@Composable
+@Preview
```
AndrewCheung360: Nit: could consider using preview parameters instead of multiple previews

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+                )
+            }
+        }
+        Spacer(modifier = Modifier.height(16.dp))
+        HighlightsFilterRow({/*todo on filter selected*/})
+    }
+}
+
+@Composable
+fun HighlightsCardRow(
```
AndrewCheung360: could be private if only used here

AndrewCheung360: Since it is a separate composable, based on consistency, it might be good to also have a preview for this

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+                Image(
+                    painter = painterResource(R.drawable.ic_right_chevron),
+                    contentDescription = "right chevron",
+                    colorFilter = ColorFilter.tint(GrayMedium, blendMode = BlendMode.SrcIn)
+                )
+            }
+        }
+        Spacer(modifier = Modifier.height(16.dp))
+        LazyRow(
+            modifier = Modifier.padding(start = 24.dp)
```
AndrewCheung360: I would check with design if we want the appearance to include the cards going into the leftmost edge of the screen when swiping to the left or if we want there to be a padding where the card disappears into. If it is the former, it will require more than just deleting this padding here because we probably want an initial space. Maybe you can add an item { spacer } to the lazy row for that. This same comment applies to the filter row as well with the sports buttons

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+                        false
+                    )
+                } else if (highlightsList[i] is ArticleHighlightData) {
+                    ArticleHighlightCard(
+                        highlightsList[i] as ArticleHighlightData,
+                        false
+                    )
+                }
+
+                Spacer(modifier = Modifier.width(16.dp))
```
AndrewCheung360: Nit: could probably use the horizontalArrangement argument and use Arrangement.spacedBy to achieve the same thing without adding extra spacer at the end

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+    modifier: Modifier = Modifier
+) {
+    Box(
+        modifier = modifier.then(
+            if (wide) Modifier.fillMaxWidth() else Modifier.width(241.dp)
+        )
+            .height(192.dp)
+            .clip(shape = RoundedCornerShape(12.dp))
+    ) {
+        AsyncImage(
```
AndrewCheung360: Same comments as above async image

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+fun VideoHighlightCardHeader(
+    image: String
+) {
+    Box(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(117.dp)
+            .clip(RoundedCornerShape(topStart = 16.dp, topEnd = 16.dp))
+    ) {
+        AsyncImage(
```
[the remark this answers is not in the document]
conniecliu: I agree ^^

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = Color.White)
+    ) {
+        Spacer(modifier = Modifier.height(24.dp))
+        HighlightsScreenHeader({})
+        Spacer(modifier = Modifier.height(24.dp))
+        if (todayHighlightsList.isEmpty() and pastThreeHighlightsList.isEmpty()) {
```
conniecliu: Ohh didn't know we could use `and` instead of `&&`. Out of curiosity, is there a difference between the two?

app/src/main/java/com/cornellappdev/score/model/Highlights.kt
```diff
@@ -0,0 +1,17 @@
+package com.cornellappdev.score.model
+
+data class VideoHighlightData(
+    val title: String,
+    val image: String,
```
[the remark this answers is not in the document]
zachseidner1: or maybe even `thumbnailImageUrl`

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+                .fillMaxSize()
+                .background(color = Color.Black.copy(alpha = 0.4f))
+        )
+    }
+}
+
+@Composable
+fun VideoHighlightCard(
+    videoHighlight: VideoHighlightData,
+    wide: Boolean,
```
zachseidner1: I think instead of having a `wide` parameter, we can just use the `modifier` parameter? The caller can just specify `modifier.fillMaxWidth()` if they want it to be wide. Lmk what you think.

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
```diff
@@ -0,0 +1,327 @@
...
+                    modifier = Modifier.weight(1f),
+                    style = heading2,
+                    text = videoHighlight.title,
+                    maxLines = 1,
+                    overflow = TextOverflow.Ellipsis
+                )
+                Row(
+                    verticalAlignment = Alignment.CenterVertically
+                ) {
+                    Image(
```
zachseidner1: You should use `Icon` for all icons

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+            ) {
+                Text(
+                    text = "${highlightsList.size} Results",
+                    style = bodyNormal
+                )
+                Image(
+                    painter = painterResource(R.drawable.ic_right_chevron),
+                    contentDescription = "right chevron",
+                    colorFilter = ColorFilter.tint(GrayMedium, blendMode = BlendMode.SrcIn)
+                )
```
zachseidner1: Use `Icon` instead, also this should be `20.dp`

app/src/main/java/com/cornellappdev/score/components/HighlightsCard.kt
amjiao: I think the LinkAnnotation.Url in BasicText handles it and automatically launches an intent!

app/src/main/java/com/cornellappdev/score/components/HighlightsFilter.kt
amjiao: oops ya, sportList is just a testing constant. i made a sportList parameter so that the actual sportList will be passed in by the view model. (Usually sportlist is a filtered list of available sports)

app/src/main/java/com/cornellappdev/score/screen/HighlightsScreen.kt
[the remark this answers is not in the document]
amjiao: fixed!

note - incomplete arguments since just UI for now, will think about navigating to a search screen and saving values to uistate in a future pr

app/src/main/java/com/cornellappdev/score/components/highlights/HighlightsSearchBar.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.cornellappdev.score.R
+import com.cornellappdev.score.theme.GrayLight
+import com.cornellappdev.score.theme.Style.bodyNormal
+
+@Composable
+fun HighlightsSearchBar(
+    onSearchClick: () -> Unit
+) {
+    val interactionSource = remember { MutableInteractionSource() }
+    var searchQuery by remember { mutableStateOf("") }
```
AndrewCheung360: You don't necessarily need to make edits here yet for this PR, but just want to mention that I think the searchQuery state might end up being hoisted up to the highlights screen composable for vm filtering logic

app/src/main/java/com/cornellappdev/score/components/highlights/VideoHighlightsCard.kt
```diff
@@ -0,0 +1,215 @@
...
+    videoHighlight: VideoHighlightData,
+    modifier: Modifier = Modifier   //for wide cards, specify Modifier.fillMaxWidth()
+) {
+    Column(
+        modifier = modifier
+            .width(241.dp)
+            .clip(RoundedCornerShape(12.dp))
+    ) {
+        VideoHighlightCardHeader(videoHighlight.thumbnailImageUrl)
+
```
AndrewCheung360: Nit: we can remove this space

app/src/main/java/com/cornellappdev/score/components/highlights/VideoHighlightsCard.kt
```diff
@@ -0,0 +1,215 @@
...
+                                    )
+                                ),
+                            )
+                        ) {
+                            append("YouTube")
+                        }
+                    }
+                )
+                Icon(
+                    painter = painterResource(R.drawable.arrow_outward_red),
```
AndrewCheung360: Not the biggest deal, but I think it might actually be possible to not need to have two separate arrows in the resources because you can probably tint the white arrow to the color you need.

app/src/main/java/com/cornellappdev/score/components/EmptyState.kt
```diff
@@ -21,58 +22,45 @@ import com.cornellappdev.score.theme.GrayPrimary
...
             .height(height)
             .fillMaxWidth(), contentAlignment = Alignment.Center
     ) {
-        EmptyState(title = title, subtitle = subtitle)
+        Column(
+            modifier = Modifier,
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center
+        ) {
+            Image(
```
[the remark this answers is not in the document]
zachseidner1: !!

app/src/main/java/com/cornellappdev/score/components/highlights/VideoHighlightsCard.kt
amjiao: ope ya that does work, noted for future usages!

PR #102, opened 2026-02-17
Fixed BaseBall Scoring and IsLive

EmilJiang:
Sometimes baseball only has 6 innings vs the normal 9. This is displayed in the result section, so I now parse that section and fix the scoreboard. Also, that causes issue in the score display, so I now parse it a bit differently. I also added the function where isLive is now accurate. 

## Summary by CodeRabbit

* **New Features**
  * Games now display time information for better scheduling
  * Live game status now determined by precise game start time instead of day-only matching

* **Bug Fixes**
  * Enhanced score parsing with improved fallback logic
  * Improved accuracy of game result calculations

* **Chores**
  * Upgraded Apollo GraphQL library to version 4.0.0


amjiao (approved):
lgtm!

PR #105, opened 2026-03-04
Calendar+ticket

EmilJiang:
Add to Calendar
Screenshot 2026-03-04 at 5 41 59 PM

Ticketing

Screenshot 2026-03-04 at 5 43 38 PM

## Summary by CodeRabbit

* **New Features**
  * Added "Buy Tickets" button to game details screen that directs users to purchase tickets
  * Added "Add to Calendar" button to enable users to add games directly to their device calendar
  * Added ticket icon to the user interface

### DIFF
--- a/app/src/main/java/com/cornellappdev/score/model/Game.kt
+++ b/app/src/main/java/com/cornellappdev/score/model/Game.kt
@@ -67,7 +67,8 @@ data class GameDetailsGame(
     val time: String?,
     val scoreBreakdown: List<List<String?>?>?,
     val team: GameDetailsTeam?,
-    val boxScore: List<GameDetailsBoxScore>?
+    val boxScore: List<GameDetailsBoxScore>?,
+    val ticketUrl: String?
 )
 
 
@@ -126,7 +127,8 @@ data class DetailsCardData(
     val daysUntilGame: Int?,
     val hoursUntilGame: Int?,
     val homeScore: Int,
-    val oppScore: Int
+    val oppScore: Int,
+    val ticketUrl: String?
 )
 
 // Scoring information by round of a game, used in the box score
@@ -146,7 +148,7 @@ data class TeamScore(
 // Aggregated game data showing scores for both teams
 data class GameData(
     val teamScores: Pair<TeamScore, TeamScore>
-){
+) {
     val maxPeriods: Int
         get() =
             maxOf(
@@ -298,7 +300,8 @@ fun GameDetailsGame.toGameCardData(): DetailsCardData {
         homeScore = convertScores(scoreBreakdown?.getOrNull(0), sport).second
             ?: parsedScores?.first ?: 0,
         oppScore = convertScores(scoreBreakdown?.getOrNull(1), sport).second
-            ?: parsedScores?.second ?: 0
+            ?: parsedScores?.second ?: 0,
+        ticketUrl = ticketUrl ?: ""
     )
 }
 
--- a/app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
+++ b/app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.content.Intent
 import android.net.Uri
 import android.provider.CalendarContract
+import android.util.Log
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.horizontalScroll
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -220,16 +230,26 @@ fun GameDetailsContent(
                     }
 
                     Spacer(modifier = Modifier.weight(1f))
```
amjiao: the spacing is slightly off here 

Image

design
Image

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -382,7 +403,37 @@ private fun EmptyGameDetailsPreview() {
...
-            oppScore = 0
-        ), navigateToGameScoreSummary = {}
+            oppScore = 0,
+            ticketUrl = ""
+        ),
+        navigateToGameScoreSummary = {},
     )
+}
+
+// helper
```
amjiao: could u move this function to the view model? just to keep logic out of the UI

amjiao (commented):
just two quick things to address, otherwise looks good!

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
EmilJiang: Screenshot 2026-04-22 at 6 05 37 PM
I couldnt find the correct measurement, but this looks kind of similar to me

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -67,7 +67,8 @@ data class GameDetailsGame(
     val time: String?,
     val scoreBreakdown: List<List<String?>?>?,
     val team: GameDetailsTeam?,
-    val boxScore: List<GameDetailsBoxScore>?
+    val boxScore: List<GameDetailsBoxScore>?,
```
caleb-bit: Nit: It's a bit confusing that this is called `boxScore` but it's a list of box scores.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -67,7 +67,8 @@ data class GameDetailsGame(
...
+    val ticketUrl: String?
```
caleb-bit: Should we use the `URI` type for this internally? I think the conversion logic should happen at the root so `Game` objects don't carry around possibly bad URLs.

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -126,7 +127,8 @@ data class DetailsCardData(
     val daysUntilGame: Int?,
     val hoursUntilGame: Int?,
     val homeScore: Int,
-    val oppScore: Int
+    val oppScore: Int,
+    val ticketUrl: String?
```
caleb-bit: same thing here

app/src/main/java/com/cornellappdev/score/screen/GameDetailsScreen.kt
```diff
@@ -211,25 +222,31 @@ fun GameDetailsContent(
...
-//                        }
-//                    )
+                    Spacer(modifier = Modifier.height(100.dp))
+                    Row() {
+                        ButtonPrimary(
+                            "Buy Tickets",
+                            painterResource(R.drawable.ticket),
+                            onClick = {
+                                gameCard.ticketUrl?.let { url ->
+                                    val intent = Intent(Intent.ACTION_VIEW, Uri.parse(url))
```
caleb-bit: This is URL parsing logic being done on the view side, and this can throw. As with previous comments, moving to model side is probably better.

app/src/main/java/com/cornellappdev/score/viewmodel/GameDetailsViewModel.kt
```diff
@@ -43,4 +51,31 @@ class GameDetailsViewModel @Inject constructor(
         applyMutation { copy(loadedState = ApiResponse.Loading) }
         scoreRepository.getGameById(gameId)
     }
+
+    fun addGameToCalendar(context: Context, gameCard: DetailsCardData) {
+        val date = gameCard.date ?: return
```
caleb-bit: We probably don't need to pass the whole `gameCard` for this since it looks like we're only using `date` and `time`. I think we should make these parameters so the function is dumber.

app/src/main/java/com/cornellappdev/score/viewmodel/GameDetailsViewModel.kt
```diff
@@ -43,4 +51,31 @@ class GameDetailsViewModel @Inject constructor(
...
+            val localTime = LocalTime.parse(time.trim().uppercase().replace(".", ""), formatter)
+            date.atTime(localTime)
+        } catch (e: Exception) {
+            Log.e("Calendar", "Failed to parse time: '$time'", e)
+            date.atStartOfDay()
+        }
+
+        val zoneId = ZoneId.systemDefault()
+        val startMillis = startDateTime.atZone(zoneId).toInstant().toEpochMilli()
+        val endMillis = startDateTime.plusHours(2).atZone(zoneId).toInstant().toEpochMilli()
```
caleb-bit: Is the 2 hours here a temporary default? What if games are longer than 2 hours for instance? Feels like there's some BE data behind this.

caleb-bit (changes requested):
Mostly looks good but left some comments.

PR #109, opened 2026-04-22
Live score

EmilJiang:
First attempt, still need to test

### DIFF
--- a/app/src/main/java/com/cornellappdev/score/model/Game.kt
+++ b/app/src/main/java/com/cornellappdev/score/model/Game.kt
@@ -302,6 +302,38 @@ fun GameDetailsGame.toGameCardData(): DetailsCardData {
     )
 }
 
+/**
+ * Merges a live socket update into the current DetailsCardData.
+ * Null fields in [update] leave existing values unchanged.
+ */
+fun DetailsCardData.applySocketUpdate(update: SocketGameUpdateData): DetailsCardData {
+    val newScoreBreakdown = update.scoreBreakdown ?: scoreBreakdown
+    val newGameData = if (update.scoreBreakdown != null) {
+        toGameData(
+            scoreBreakdown = newScoreBreakdown,
+            team1 = TeamBoxScore("Cornell"),
+            team2 = TeamBoxScore(opponent),
+            sport = sport
+        )
+    } else gameData
+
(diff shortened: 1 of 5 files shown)


caleb-bit:
@coderabbitai review

app/src/main/java/com/cornellappdev/score/model/SocketModels.kt
```diff
@@ -0,0 +1,37 @@
...
+    val time: String? = null,
+    val description: String? = null,
+    val scorer: String? = null,
+    val assist: String? = null,
+    val scoreBy: String? = null,
+    val corScore: Int? = null,
+    val oppScore: Int? = null
+)
+
+fun SocketBoxScoreEntry.toGameDetailsBoxScore() = GameDetailsBoxScore(
```
caleb-bit: Is there a reason we need to make the same class? Why not just make `GameDetailsBoxScore` serializable? If there is a semantic reason that's fine

app/src/main/java/com/cornellappdev/score/model/Game.kt
```diff
@@ -304,7 +305,41 @@ fun GameDetailsGame.toGameCardData(): DetailsCardData {
...
+/**
+ * Merges a live socket update into the current DetailsCardData.
+ * Null fields in [update] leave existing values unchanged.
+ */
+fun DetailsCardData.applySocketUpdate(update: SocketGameUpdateData): DetailsCardData {
+    val newScoreBreakdown = update.scoreBreakdown ?: scoreBreakdown
+    val newGameData = if (update.scoreBreakdown != null) {
+        toGameData(
+            scoreBreakdown = newScoreBreakdown,
+            team1 = TeamBoxScore("Cornell"),
```
caleb-bit: We should have a constant for `"Cornell"`.

caleb-bit (commented):
Looks good overall, just left some nits/questions.

app/src/main/java/com/cornellappdev/score/model/Game.kt
EmilJiang: good point

app/src/main/java/com/cornellappdev/score/model/SocketModels.kt
EmilJiang: done

